PR #2, opened 2015-12-14
Adding decode token methods

minglis:
Implements token checking …
- all JWT code wrapped in library
- clients can construct and check tokens

Token checking:

Checks are as follows:
- signature
- expiry time (not in the future, not too old)
- request (METHOD / PATH)
- payload (optional)

Expiry time checking …
- JWT library already checks whether token is in the future so removed this check.

### DIFF
--- a/client/__init__.py
+++ b/client/__init__.py
@@ -1,4 +1,5 @@
-from .errors import APIError, HTTPError, InvalidResponse
+from .errors import APIError, HTTPError, InvalidResponse, TokenError, \
+    TokenDecodeError, TokenExpiredError, TokenPayloadError, TokenRequestError
 from .errors import REQUEST_ERROR_STATUS_CODE, REQUEST_ERROR_MESSAGE
 
 from .notifications import NotificationsAPIClient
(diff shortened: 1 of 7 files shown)


client/__init__.py
```diff
@@ -1,4 +1,5 @@
-from .errors import APIError, HTTPError, InvalidResponse
+from .errors import APIError, HTTPError, InvalidResponse, TokenError, \
```
servingUpAces: I would prefer if we used absolute paths on the imports.

client/jwt.py
```diff
@@ -51,28 +56,112 @@ def create_jwt_token(request_method, request_path, secret_key, client_id, reques
...
+    :param request_method: HTTP method for the request
+    :param request_path: Resource path for the request
+    :param request_payload: Body of the request
+    :return boolean: True if valid token, False otherwise
+    :raises AssertionError: If any required fields are not present
+    :raises jwt.DecodeError: If signature validation fails
+    """
+    try:
+        # check signature of the token
+        decoded_token = jwt.decode(token, key=secret.encode(), verify=True, algorithms=[__algorithm__])
```
servingUpAces: Is the encode alright without a charset?

PR #13, opened 2016-02-23
Adding email to utils

minglis:
Notifications client send email method updated 
- signature matches that of SMS
- takes email address and template id

- Removed the send email with content endpoint


minglis:
Note this is a breaking change, so once merged I'll need to ensure the new version works on admin - so I'll do an admin release after merge to bring this in.

notifications_python_client/notifications.py
```diff
@@ -2,28 +2,20 @@
...
-        if template_id:
-            notification.update({
-                "template": template_id
-            })
-        if content:
-            notification.update({
-                "content": content
-            })
+    def send_sms_notification(self, to, template_id=None):
+        notification = {
```
servingUpAces: Would the template_id ever be none?

minglis: Fixed - missed this when removing content option

PR #16, opened 2016-03-10
Logger calls added replace print statements.

mannickutd:


quis:
There’s a logging thing in utils, which is used by the admin app: https://github.com/alphagov/notifications-utils/blob/master/utils/logging.py

I _think_ it logs as JSON, which is easier to search/filter in Cloudwatch. Could be worth using this instead/asking @minglis ?

minglis:
Is it possible to use the logger from the parent apps in here? not sure how this logs currently, but it would be good if it used the app/line/etc from the parent app and the JSON log formatter as per the flask apps.

servingUpAces:
:+1:

PR #17, opened 2016-03-11
Add personalisation to Python client

quis:
The API accepts personalisation of messages since:
https://github.com/alphagov/notifications-api/pull/109

This commit add `personalisation` as an optional parameter when sending emails and text messages.

It also makes the assertions in the tests a bit more specific.

Minor version bump, adding a new feature in a non-breaking way.


README.md
```diff
@@ -64,16 +64,19 @@ Creating a text message:
...
-    
-* "template_id" is the template to send
+
+* `template_id` is the template to send
     * Must be an integer that identifies a valid template. Templates are created in the admin tools.
-    
+
+* `personalisation` is the template to send
+    * Must be an dictionary, with keys matching the placeholders in the template.
+
```
servingUpAces: Let's add a sample dict. I got an error when creating one with single quotes.

servingUpAces:
Just a small update to the README, otherwise :+1:

README.md
quis: Updated the README in d10a943. It needs to be valid JSON, not just any old dict (which is why the single quotes didn’t work).

PR #18, opened 2016-03-29
Added get_all_notifications to client

minglis:
- new call to fetch all notifications based on service
- handles filter params (status/template type)
- Utils script updated to handle new call

### DIFF
--- a/notifications_python_client/notifications.py
+++ b/notifications_python_client/notifications.py
@@ -26,3 +26,18 @@ class NotificationsAPIClient(BaseAPIClient):
 
     def get_notification_by_id(self, id):
         return self.get('/notifications/{}'.format(id))
+
+    def get_all_notifications(self, status=None, template_type=None):
+        data = {}
+        if status:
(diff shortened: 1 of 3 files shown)


notifications_python_client/notifications.py
```diff
@@ -26,3 +26,18 @@ def send_email_notification(self, to, template_id, personalisation=None):
...
+    def get_all_notifications(self, status=None, template_type=None):
+        data = {}
+        if status:
+            data.update({
+                'status': status
+            })
+        if template_type:
+            data.update({
+                'template_type': template_type
+            })
```
quis: `data = {key: eval(key) for key in ['status', 'template_type'] if eval(key)}` :wink:

quis:
Tested locally and :+1:, with or without the one-liner

notifications_python_client/notifications.py
minglis: vomits over his shoes

PR #19, opened 2016-04-14
Simplify jwt

servingUpAces:
This pull request removes the need for the req and pay in the Json Web Token. This should simplify the request to the notifications-api application. 
This version will continue to take in the request and payload to the create_jwt_token and decode_jwt_token making this version backward compatible for other clients.
Once the API and other clients and documentation is updated the methods can be updated to remove these as parameters.

https://www.pivotaltracker.com/story/show/116971293


quis:
Seen demo, code looks good, :+1:

PR #20, opened 2016-05-04
Remove request parameters from authentication methods

servingUpAces:
Remove request_method, request_path and request_body from authentication decode and encode methods.

Change version number to 1.0.0; this is a breaking change.


notifications_python_client/authentication.py
```diff
@@ -27,11 +27,8 @@ def create_jwt_token(request_method, request_path, secret, client_id, request_bo
     req: signed request, of the format METHOD PATH. Example POST /resource
```
minglis: Do these need removed? mentions to removed fields.

PR #21, opened 2016-05-20
Notification statistics endpoint

leohemsted:
See https://github.com/alphagov/notifications-api/pull/333 for endpoint details


quis:
This looks 👌. Should include a version bump as well.

quis:
👍

PR #22, opened 2016-06-23
Rewrite the documentation as a get-started guide

quis:
#### [See it here](https://github.com/alphagov/notifications-python-client/blob/4c7276e5270c819e91a360034cff3a74ef15b342/README.md)

The `README`s for our clients should help you get set up making API calls to Notify quickly and easily. Therefore you don’t need to know about how to run the clients tests, or develop on it locally.

This commit rewrites the README to contain just the bare minimum you’d need to integrate the client into a new Python project.

This is sort of a strawperson that we can talk about.

The stuff specific to working on the client has moved into `CONTRIBUTING.md`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,50 @@
+# Contributing
+
+Pull requests welcome.
+
+## Working on the client locally
+
+This is a Python 3 application. It has not been tested against any
+version of Python 2.
+
+```shell
+    brew install python3
+```
+
+## Dependency management
+
+This is done through [pip](pip.readthedocs.org/) and
+[virtualenv](https://virtualenv.readthedocs.org/en/latest/).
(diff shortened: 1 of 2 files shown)


leohemsted:
looks good, but do you want to talk about api return values? (especially since for example this client returns json vs java client returning objects)

quis:
Had a chat with @catherineheywood. I think that for now we should:

- talk about what kind of thing you get back (eg, JSON, a `HTTPError()`) rather than the contents of it
- link to the API doc for the whole response

Will do that shortly.

quis:
@leohemsted Added documentation of responses/exceptions in 0fb5374.

quis:
Merging this now so we have it in time for the research run through this afternoon.

leohemsted:
i give this a retrospective 👍

PR #24, opened 2016-07-26
Add helpful string representation of exceptions

quis:
It’s hard to debug what’s going on with the API when you have to know the structure of the `Exception` object that gets returned by the client.

This commit adds a `__str__` method to the exceptions that the client raises. This means that the error code and message will be displayed in any tracebacks that a developer might be looking at.

screen shot 2016-07-26 at 10 32 08
screen shot 2016-07-26 at 10 45 46

***

Bumps patch version to 1.1.1 and removes the version from the README so that developers always get the latest version when they’re starting out.


leohemsted:
👍

PR #26, opened 2016-08-09
DO NOT MERGE: Can we use details/summary in Markdown?

quis:
Testing details/summary in Markdown


quis:
@catherineheywood This seems to work, see: https://github.com/alphagov/notifications-python-client/blob/5846cc29bbe2a95b32aa21aea30e05255f32151f/README.md#get-the-status-of-one-message

PR #27, opened 2016-08-11
Responses

catherineheywood:
Include responses in Python client readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,20 @@ If a template has placeholders, you need to provide their values.
         </td>
     </tr>
     <tr>
-        <td>201
+        <td>429
+        </td>
+        <td>{"result": "error", <br> "message": "Exceeded send limits (50) for today"}
+        </td>
+    </tr>
+        <tr>
+        <td>400
+        </td>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -62,34 +55,176 @@ notifications_client.send_sms_notification(
...
+        </td>
+    </tr>
+        </tr>
+        <tr>
+        <td>400
+        </td>
+        <td>
+<pre>
+{"result":"error", 
+"message"={"to": ["Invalid {notification_type} for restricted service")]}
```
leohemsted: we've changed these messages - there's now a couple of messages they might see instead (still with 400 response code):

`"Can’t send to this recipient using a team-only API key"`

or if they're in trial mode 

`"Can’t send to this recipient when service is in trial mode – see https://www.notifications.service.gov.uk/trial-mode"`

quis:
Can whoever merges this squash all the commits into one pls. Thnx.

leohemsted:
superceded by https://github.com/alphagov/notifications-python-client/pull/28

PR #29, opened 2016-09-08
Make `init` parameters match what we call them

quis:
We call the things that the clients need the ‘service ID’ and the ‘API key’. The client calls them `client_id` and `secret`. This is confusing if you’re a user of the client because you start by referring to them by the names the admin app gives them.

The terms used in the UI are more readily understood, as evidenced by the amount of research we’ve done with them in place.

This commit makes two changes:

1. Rename the arguments that the client’s `init` method takes to match the terms we use in the UI.

2. Update the documentation to use keyword arguments when initialising the client. For someone looking at the code down the line `NotificationsClient(service_id='5c85a3d2-0ab9-44de-bb9b-45eedf2ab831')` is more helpful than `NotificationsClient('5c85a3d2-0ab9-44de-bb9b-45eedf2ab831')`, when both the `service_id` and `api_key` are UUIDs.

This is a breaking change (and is versioned as such) because it renames a method’s arguments. The clients that inherit from this one will also need their method arguments and `Super` calls updating, eg https://github.com/alphagov/notifications-admin/blob/ae203dfe53bbc167b2146024b6513b391fc8982e/app/notify_client/status_api_client.py#L5


notifications_python_client/base.py
```diff
@@ -27,11 +27,11 @@ def __init__(self, base_url=None, client_id=None, secret=None):
         :return:
         """
         assert base_url, "Missing base url"
-        assert client_id, "Missing client id"
-        assert secret, "Missing secret"
+        assert service_id, "Missing client id"
+        assert api_key, "Missing secret"
```
leohemsted: should change the assert messages too

quis: Good point, fixed and squashed.

leohemsted:
👍

PR #32, opened 2016-09-16
Added include jobs to python client

minglis:
- new param, true or false
- by default no CSV notifications returned unless requested.


tests/notifications_python_client/test_notifications_api_client.py
```diff
@@ -14,22 +14,35 @@ def test_get_notification_by_id(notifications_client, rmock):
     assert rmock.called
 
 
-def test_get_all_notifications_by_stype_and_status(notifications_client, rmock):
-    endpoint = "{0}/notifications?status={1}&template_type={2}".format(TEST_HOST, "status", "type")
-    rmock.request(
+def test_get_all_notifications_by_stype_and_status_and_jobs(notifications_client, rmock):
```
quis: Typo: `stype`

tests/notifications_python_client/test_notifications_api_client.py
```diff
@@ -14,22 +14,35 @@ def test_get_notification_by_id(notifications_client, rmock):
...
+    endpoint = "{0}/notifications?status={1}&template_type={2}".format(TEST_HOST, "status", "type", True)
```
quis: There’s no `{3}` in the string here. You don’t need to pass in a 4<sup>th</sup> parameter to the `format()`.

minglis:
Not needed. Added only for functional tests, which don't use the client at this level.

PR #36, opened 2016-09-27
Make secret include API key and service ID

quis:
In research we’ve seen people mix up the service ID and API key because they’re both 36 character UUIDs. We can’t get rid of the service ID because it’s used to look up the API key.

Instead, we should change API key to be one long string, which contains both the service ID, API key and (optionally) the name of the key. For example:

**API key**
```
casework_production-8b3aa916-ec82-434e-b0c5-d5d9b371d6a3-dcdc5083-2fee-4fba-8afd-51f3f4bcb7b0
```

This commit changes the client to make the service ID optional. If it’s not passed, they we attempt to extract it from the `secret` instead.

If we decided to go with this then it would mean rolling out this change to all the clients first, then changing the admin app to print out the keys in the new format.


quis:
@minglis made it a bit more robust based on your feedback, see 136911d and 6317ce3.

tests/notifications_python_client/test_base_api_client.py
```diff
@@ -1,8 +1,62 @@
...
+            '-'
+            '8b3aa916-ec82-434e-b0c5-d5d9b371d6a3'  # secret
+        )
+    ),
+    BaseAPIClient(
+        api_key=(
+            'name_of_key'
+            '😬'
+            'c745a8d8-b48a-4b0d-96e5-dbea0165ebd1'  # service ID
+            '😬'
```
leohemsted: 👏

PR #39, opened 2016-10-18
Instantiate client with combined API key in readme

quis:
So that users don’t have to worry about:
- base URL
- service ID

Depends on:
- [x] https://github.com/alphagov/notifications-admin/pull/968


quis:
@leohemsted 4af8104

PR #40, opened 2016-11-03
remove AssertionErrors

leohemsted:
replace AssertionErrors with new TokenDecodeError subclasses

if your provided JWT token is missing either the 'iss' or 'iat' fields,
we now raise TokenIssuerError or TokenIssuedAtError respectively,
rather than AssertionErrors


leohemsted:
@quis iat test is found in admin app, since i wanted to make sure i was also capturing our authhandler's logic

https://github.com/alphagov/notifications-api/pull/726/commits/f089b7512915d4b8f5724f33acd1f248d204c27d#diff-707b2a6eab3566cfd44abdb93239c959R67

PR #41, opened 2016-11-07
allow api_key to be positional

leohemsted:
allow user to say `NotificationAPIClient(foo)` instead of `...(api_key=foo)` (since thats what we say in our docs, and it makes sense)

also removed monotonic as dependency.

this is a *breaking* change

- [x] check our usage to make sure we won't break any of our usage by doing this
  * [x] admin
  * [x] api
  * [x] functional tests
    + some changes required
  * [x] py client
- [x] ~~check if any other teams are using python client and if this would break their usage~~
  * not going to do. added changelog.md so peeps can see if it's worth upgrading
- [x] resolve merge conflicts and bump version again


matthew-shaw:
Thanks! 👍 ⭐️

PR #45, opened 2016-12-05
publish the python client to pypi

leohemsted:
## setup.py work

`python setup.py test`
`python setup.py integration_test`

note: these methods will download eggs of requirements and store them locally (in `.eggs` folder) rather than using reqs from your python path, so don't recommend using them in regular development - stick with `make test` or `scripts/run_tests.sh` or whatever you use.

## makefile work

see https://github.com/alphagov/notifications-python-client/pull/45/commits/faca7efdca0d60d1c60f06d2d9322e71418d1449 for details - added `make publish-to-pypi[-with-docker]`, which are fairly self explanatory. You shouldn't actually be running these locally though (unless you're publishing to testpypi to test some change to packaging/docs) - lets only run them from jenkins 

# developer experience

```
(pypi-test) [leohemsted:~/dev/python-client]$ make publish-to-pypi-with-docker
[ ... ]
Uploading distributions to https://testpypi.python.org/pypi
Uploading notifications_python_client-3.1.3-py3-none-any.whl
```

# user experience

```
[leohemsted:~]$ mkvirtualenv pypi-test -p python3                                                                                                                                 
[ ... ]
(pypi-test) [leohemsted:~]$ pip install notifications-python-client -i https://testpypi.python.org/pypi
[ ... ]
(pypi-test) [leohemsted:~]$ python
Python 3.5.2 (default, Sep 28 2016, 18:08:09)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.38)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import notifications_python_client
>>> notifications_python_client.version.__version__
'3.1.0'
```


Makefile
```diff
@@ -36,6 +42,21 @@ test: venv ## Run tests
 integration-test: ## Run integration tests
 	./scripts/run_integration_tests.sh
 
+.PHONY: build-wheel
+build-wheel: venv ## build distributable wheel
+	pip install wheel
```
bandesz: Should this be installed in the venv? (./venv/bin/pip install wheel). The question stands for the python/twine runs too.

Makefile
```diff
@@ -85,9 +106,25 @@ integration-test-with-docker: prepare-docker-runner-image generate-env-file ## R
...
+publish-to-pypi-with-docker: prepare-docker-runner-image generate-env-file ## publish wheel to pypi inside a docker container
+	docker run -i --rm \
+		--name "${DOCKER_CONTAINER_PREFIX}-publish-to-pypi" \
+		-v `pwd`:/var/project \
+		-e http_proxy="${HTTP_PROXY}" \
+		-e HTTP_PROXY="${HTTP_PROXY}" \
+		-e https_proxy="${HTTPS_PROXY}" \
+		-e HTTPS_PROXY="${HTTPS_PROXY}" \
+		-e NO_PROXY="${NO_PROXY}" \
+		-e PYPI_USERNAME="${PYPI_USERNAME}" \
```
bandesz: we should pass in PYPI_REPOSITORY_NAME and PYPI_REPOSITORY_URL too

leohemsted: even though it's defined at the top of this file, and we're only running a command from this file inside the docker?

bandesz: if we would ever pass this in from Jenkins, the values would be still the default ones inside Docker

Makefile
```diff
@@ -36,6 +42,21 @@ test: venv ## Run tests
...
 
+.PHONY: build-wheel
+build-wheel: venv ## build distributable wheel
+	./venv/bin/pip install wheel
+	./venv/bin/python setup.py bdist_wheel
+
+.PHONY: publish-to-pypi
+publish-to-pypi: build-wheel ## upload distributable wheel to pypi
+	./venv/bin/pip install twine
+	twine upload dist/*.whl \
```
bandesz: ./venv/bin/twine here too

leohemsted: ah fuck

setup.py
```diff
@@ -1,23 +1,50 @@
...
+
+
+install_requires = get_reqs('requirements.txt')
+tests_require = get_reqs('requirements_for_test.txt')
+# for running pytest as `python setup.py test`, see
+# http://doc.pytest.org/en/latest/goodpractices.html#integrating-with-setuptools-python-setup-py-test-pytest-runner
+setup_requires = ['pytest-runner']
+
+
+class IntegrationTestCommand(TestCommand):
```
imdadahad: Minor point but would be nice put this in a separate file, e.g. scripts/integration_test_command.py and then import and use in cmdclass. Just keeps the file lean.

leohemsted: possibly. all the examples i saw have it in setup.py ( and not just tutorials which might be kept together for brevity, eg https://github.com/pytest-dev/pytest/blob/master/setup.py#L86, https://github.com/zzzeek/sqlalchemy/blob/master/setup.py#L77), because it makes no sense outside of the setup process. it's not part of the actual usage of the lib or its surrounding helpers if that makes sense - all it's doing is calling `py.test integration_tests/integration_test_suite.py` anyway

imdadahad: cool 👍

PR #46, opened 2016-12-07
Import monotonic from monotonic

minglis:


leohemsted:
i was talking to andras - once the pypi work's done we need to get the client supporting 2.7 properly - actually running the tests through both versions, uploading the wheels to pypi etc.

PR #49, opened 2016-12-15
Upgrade to version of the API

servingUpAces:
This version of the python client uses version 2 of the api. Please refer to the README for all details about the requests and repsonse
- change to the send_sms response
- change to the send_email response
- change to the get_notificaiton_by_id response
- change to the get_notificaitons response
- new filter options for get_notificaitons
- better error messages
- update README to reflect the changes.
Now at version 4.0.0


integration_test/schemas/v2/notification_schemas.py
```diff
@@ -0,0 +1,220 @@
...
+        }},
+        {"properties": {
+            "line_1": {"type": "string", "minLength": 1},
+            "postcode": {"type": "string", "minLength": 1},
+            "type": {"enum": ["letter"]},
+
+            "email_address": {"type": "null"},
+            "phone_number": {"type": "null"}
+        }}
+    ],
```
leohemsted: does this `oneOf` block work in conjunction with the properties block below? ie: does it actually moan if both email and phone are supplied or if the type is wrong

CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
...
+* Update to `NotificationsAPIClient.send_sms_notification()`: 
+    * added `reference`: an optional unique identifier for the notification or an identifier for a batch of notifications. `reference` can be omitted.
+    * the `dict` returned has changed, learn more from the README. 
+* Update to `NotificationClient.send_email_notification()`: 
+    * added `reference`: an optional unique identifier for the notification or an identifier for a batch of notifications. `reference` can be omitted.
+    * the `dict` returned has changed, learn more from the README. 
+* Update to the response of `NotificationsAPIClient.get_notification_by_id(id)`, see the README for details.
+* `NotificationsAPIClient.get_all_notifications()` 
+    * Notifications can be now be filtered by `reference`, see the README for details.
+    * NotificationList response has changed, see the README for details.
```
leohemsted: should we remove `service_id` from  BaseAPIClient (and mention it here?) like we've done in the other languages?

integration_test/schemas/v2/notification_schemas.py
servingUpAces: `oneOf` does not work with the required section. 
We made a choice that the response always looks the same, meaning we have both phone_number and email_address in the schema, but only one of them should be

PR #51, opened 2017-01-24
Python 2 support

leohemsted:
🙌 🙌 🙌 🙌 🙌 

We now officially support python 2! Run tests on both 2.7 and 3.4 to ensure conformity

Biggest change is rather than the docker file stemming from python3.4:slim, it now just goes from debian:jessie (python3.4:slim inherits from that), and then apt-gets both python 2 and python 3, and then we use tox.

You can run tox locally by installing it in your env of choice (it doesn't matter, as when you use tox it'll create its own venvs for its tests), and then simply running `tox`, or it runs in docker through the command `make tox-with-docker`

Todo:

- [x] ensure we can build wheels that have the correct python 2.7 compat flags
- [ ] [AFTER MERGE] update jenkins job to run `make tox-with-docker` rather than `make test-with-docker`
- [x] [AFTER MERGE] remove contents of all python-client workspaces and rebuild, so that it picks up docker changes (PR one already done, need to update master / client tests)

### DIFF
--- a/notifications_python_client/base.py
+++ b/notifications_python_client/base.py
@@ -1,18 +1,20 @@
 from __future__ import absolute_import
+from __future__ import unicode_literals
+from __future__ import print_function
+from __future__ import division
+from future import standard_library
+standard_library.install_aliases()
+import urllib.parse
 import logging
 import json
+
 from monotonic import monotonic
+import requests
 
 from notifications_python_client.version import __version__
 from notifications_python_client.errors import HTTPError, InvalidResponse
 from notifications_python_client.authentication import create_jwt_token
 
-try:
-    import urlparse
-except ImportError:
-    import urllib.parse as urlparse
-
-import requests
 
 logger = logging.getLogger(__name__)
 
@@ -30,7 +32,6 @@ class BaseAPIClient(object):
         :param secret - application secret - used to sign the request:
         :return:
         """
-
         service_id = api_key[-73:-37]
         api_key = api_key[-36:]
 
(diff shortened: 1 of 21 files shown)


tox.ini
```diff
@@ -0,0 +1,23 @@
...
+# can't use the scripts directly as they set up the venv dir, and i'm not sure how that'll interact with tox's internal venv
+commands =
+  {envbindir}/pep8 --exclude=venv,.tox .
+  py.test tests/ {posargs}
+  {envbindir}/python integration_test/integration_tests.py
+
+# environment variables used by the tests
+passenv =
+  NOTIFY_API_URL
+  SERVICE_ID
```
imdadahad: You can remove this as we're using new api key format.

setup.cfg
```diff
@@ -1,10 +1,16 @@
...
 
 [aliases]
 test=pytest
 
 [tool:pytest]
 addopts = --verbose
 python_files = tests/**
+
+[bdist_wheel]
+# this flags the wheel as being python python 2 and 3 compatible
```
imdadahad: `python python`, sorry being super pedantic here.

notifications_python_client/base.py
```diff
@@ -71,7 +72,7 @@ def request(self, method, url, data=None, params=None):
             self.service_id
         )
 
-        url = urlparse.urljoin(self.base_url, url)
+        url = urllib.parse.urljoin(str(self.base_url), str(url))
```
imdadahad: Nice 👍

leohemsted: urlparse is the python 3 version of urllib.parse - however the future thing patches that into old urllib rather than vice versa :(

unless nice about casting everything to `str`

i hate unicode

imdadahad: 😆

PR #52, opened 2017-01-25
Add feature to get all notifications 

imdadahad:
This adds the capability for the client to retrieve all notifications using an iterator. 

## Existing setup

Currently when we get all notifications, they are paginated so only 50 are returned. In the response we do include a next link that will retrieve the next set of notifications (if they're are any). However we do not provide a way for the client to actually use this link.

## New setup

A new get all notifications iterator has been added that allows the client to retrieve notifications using an iterator. The end user does not need to worry about processing the next link, this does it for them and will simply return back a generator that they can consume to get _more_ notifications.

This also constitutes a minor version bump.


imdadahad:
Made a slight mess of this, will re-submit in a separate pr

PR #54, opened 2017-02-03
Add get all notifications iterator

imdadahad:
This adds the capability for the client to retrieve all notifications using an iterator. 

## Existing setup

Currently when we get all notifications, they are paginated so only 50 are returned. In the response we do include a next link that will retrieve the next set of notifications (if they're are any). However we do not provide a way for the client to actually use this link.

## New setup

A new get all notifications iterator has been added that allows the client to retrieve notifications using an iterator. The end user does not need to worry about processing the next link, this does it for them and will simply return back a generator that they can consume to get _more_ notifications.

## How to test

First ensure that you have some notifications sent via the API. You can override the `API_PAGE_SIZE` in the `config.py` to be a small number and then test that multiple calls are made to the api once you call the `get_all_notifications_iterator` method. 

This also constitutes a minor version bump.


notifications_python_client/notifications.py
```diff
@@ -50,6 +54,26 @@ def get_all_notifications(self, status=None, template_type=None, reference=None)
             params=data
         )
 
+    def get_all_notifications_iterator(self, status=None, template_type=None, reference=None, older_than=None):
+        result = self.get_all_notifications(status, template_type, reference, older_than)
+        notifications = result.get('notifications')
+        for notification in notifications:
+            yield notification
+
+        import re
```
leohemsted: import at top of file

notifications_python_client/notifications.py
```diff
@@ -50,6 +54,26 @@ def get_all_notifications(self, status=None, template_type=None, reference=None)
             params=data
         )
 
+    def get_all_notifications_iterator(self, status=None, template_type=None, reference=None, older_than=None):
+        result = self.get_all_notifications(status, template_type, reference, older_than)
```
servingUpAces: older_than parameter is missing from the get_all_notifications method

leohemsted:
oh and i forgot to say earlier, this absolutely needs tests.

notifications_python_client/notifications.py
```diff
@@ -50,6 +54,26 @@ def get_all_notifications(self, status=None, template_type=None, reference=None)
             params=data
         )
 
+    def get_all_notifications_iterator(self, status=None, template_type=None, reference=None, older_than=None):
```
imdadahad: Since this is literally a wrapper around `get_all_notifications`, the logic for error handling is delegated there. In other words this doesn't introduce any new error handling. `get_all_notifications` does raise on an API error however.

PR #63, opened 2017-04-19
V2 get all templates

kentsanggds:
Updated python client to use v2 get all templates and updated documentation, unit and integration tests.


leohemsted:
need to bump the version (one minor version for new backwards compat features)

also, need to update the changelog, and it's a bit out of date too... something like...

```
## 4.3.0

###Changed
* [briefly explain template API]

## 4.2.0

### Changed
* `NotificationsAPIClient.get_all_notifications()` now

## 4.1.0

### Changed
* Python client now compatible with python27

## 4.0.0

### Changed
[...]
```

README.md
```diff
@@ -630,6 +630,51 @@ Otherwise the client will raise a `HTTPError`:
 </table>
 </details>
 
+## Get all templates
+
+```python
+response = notifications_client.get_all_templates(
+    template_type=None # optional - SMS_TYPE | EMAIL_TYPE
```
leohemsted: we should reference the valid types here in this doc - think this should say `# optional - see https://github.com/alphagov/notifications-python-client/blob/master/README.md#template_type` or something. or actually, if you include that outside of the code block, then it can be a nice hyperlink

kentsanggds: ok, good call, will create another section

kentsanggds: Will have a stab at updating the version documentation too, seems like a while since that was done

PR #71, opened 2017-08-30
Add Letters API to Python Docs

jonathanglassman:
Send Messages section:
- Amended arguments to be channel-specific and copied into each channel section from the end of the Send Messages section
- Added phone number argument to text message channel section

Get Status... sections:
- Updated statuses for Letters to created, sending, failed, technical failure
- Amended failed statuses to be channel appropriate

Corrections:
- Added note saying that we support up to six lines and a postcode and this will be put in the address block of the letter
- Changed all address line 2 for letters to be required rather than optional
- Changed all fake addresses to 123 High street

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,38 @@ Otherwise the client will raise a `HTTPError`:
 </table>
 </details>
 
+<details>
+<summary>
+Arguments
+</summary>
+
+#### `phone_number`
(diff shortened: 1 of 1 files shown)


leohemsted:
Looks good, thanks @jonathanglassman!

PR #78, opened 2017-10-05
Update post email with an optional email reply to

kentsanggds:
Updated integration and unit tests for python client

Reference - 
Step 6 from this story https://www.pivotaltracker.com/story/show/150440099

NOTE - Requires the email_reply_to_id API endpoint to be available before merging

### DIFF
--- a/scripts/generate_docker_env.sh
+++ b/scripts/generate_docker_env.sh
@@ -17,6 +17,7 @@ env_vars=(
     EMAIL_TEMPLATE_ID
     SMS_TEMPLATE_ID
     LETTER_TEMPLATE_ID
+    EMAIL_REPLY_TO_ID
 )
 
 for env_var in "${env_vars[@]}"; do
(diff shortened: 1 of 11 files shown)


scripts/generate_docker_env.sh
```diff
@@ -17,6 +17,7 @@ env_vars=(
     EMAIL_TEMPLATE_ID
     SMS_TEMPLATE_ID
     LETTER_TEMPLATE_ID
+    EMAIL_REPLY_TO_ID
```
sakisv: You also need to add this to `tox.ini`

klssmith:
Looks good - could document `email_reply_to_id` under the arguments section [here](https://github.com/alphagov/notifications-python-client/blob/master/README.md#email) in the README.md too.

And maybe the `make_api_call.py` script could prompt for `email_reply_to_id` when you use the `create_email_notification` function.

scripts/generate_docker_env.sh
kentsanggds: done

PR #82, opened 2017-11-01
Allow send_sms_notification to accept an sms_sender_id

klssmith:
When using the `send_sms_notification` function, you can now choose to specify which
sender you want the notification to come from by passing in an `sms_sender_id`. If this 
is not included, the sms notification will come from the default sender.

[Pivotal story](https://www.pivotaltracker.com/story/show/152106587)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 4.6.0
+
+### Changed
+
+* Update to `NotificationsAPIClient.send_sms_notification()`
+    * added `sms_sender_id`: an optional `sms_sender_id` specified when adding SMS senders under service settings. If this is not provided, the SMS sender will be the service default SMS sender. `sms_sender_id` can be omitted.
+
 ## 4.5.0
 
 * Update to `NotificationsAPIClient.send_email_notification()`
--- a/integration_test/integration_tests.py
+++ b/integration_test/integration_tests.py
@@ -22,14 +22,16 @@ def validate(json_to_validate, schema):
     validator.validate(json_to_validate, schema)
 
 
-def send_sms_notification_test_response(python_client):
+def send_sms_notification_test_response(python_client, sender_id=None):
     mobile_number = os.environ['FUNCTIONAL_TEST_NUMBER']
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 4.6.0
+
+### Changed
+
+* Update to `NotificationsAPIClient.send_sms_notification()`
```
richardc0: Could match the email text below and substitute email => sms? But if the text is approved then its good.

tests/notifications_python_client/test_notifications_api_client.py
```diff
@@ -132,6 +132,23 @@ def test_create_sms_notification_with_personalisation(notifications_client, rmoc
     }
 
 
+def test_create_sms_notification_with_sms_sender_id(notifications_client, rmock):
```
richardc0: Add a test to test for an invalid sender_id?

integration_test/integration_tests.py
```diff
@@ -22,14 +22,16 @@ def validate(json_to_validate, schema):
     validator.validate(json_to_validate, schema)
 
 
-def send_sms_notification_test_response(python_client):
+def send_sms_notification_test_response(python_client, sender_id=None):
```
richardc0: Add a new test for sender_id being populated.
All of the previous tests shouldn't need updating as the sender_id is optional?

CHANGELOG.md
klssmith: I've changed the text to match here

integration_test/integration_tests.py
klssmith: This bit here is being used as a helper function for the tests. The test for sending an SMS without a sender id has stayed the same, and lines 157 and 167 now use this function to test sending an SMS with the sender id populated.

tests/notifications_python_client/test_notifications_api_client.py
klssmith: Leaving this as discussed to keep things consistent with the rest of the code and the tests in the other clients

PR #86, opened 2017-11-23
Variable length lists of values

jonathanglassman:
Added content stating that you can pass variable length lists of values into the personalisation field.

The client library does not currently explicitly state this.

@leohemsted will check syntax and if this applies to other languages.

## Checklist

  - `README.md`
  - `notifications_python_client/__init__.py`
  - `notifications_python_client/test_base_api_client.py`, function `test_user_agent_is_set`
  - `notifications-python-client/scripts/generate_docker_env.sh`
  - `notifications-python-client/tox.ini`
  - `CONTRIBUTING.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,12 +116,12 @@ You can omit this argument if you do not require a reference for the notificatio
 
 ##### `personalisation`
 
-If a template has placeholders, you need to provide their values, for example:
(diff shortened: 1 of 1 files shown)


quis:
I think this is confusing because it changes the primary example from the normal use case (a string) to a very niche use case (a list).

PR #98, opened 2018-02-01
move from pep8 to flake8

leohemsted:


notifications_python_client/__init__.py
```diff
@@ -15,8 +15,8 @@
 #
 # -- http://semver.org/
 
-__version__ = '4.7.1'
+__version__ = '4.7.2'
 
-from notifications_python_client.errors import REQUEST_ERROR_STATUS_CODE, REQUEST_ERROR_MESSAGE
+from notifications_python_client.errors import REQUEST_ERROR_STATUS_CODE, REQUEST_ERROR_MESSAGE  # noqa
```
leohemsted: not sure if we want these lines here? these appear to be imported here to deliberately expose to clients so i don't really want to remove htem

PR #104, opened 2018-04-18
dont use pip internals to try and read reqs.txt

leohemsted:
importing pip and using its internals to parse requirements.txt files
isn't supported or encouraged - and in v10.0.0 the internal api broke.

So, overhaul of requirements parsing.

* install requirements are now in setup.py (tox picks em up automagically)
* test requirements are now listed in tox
* flake8 is now its own separate tox command


tox.ini
```diff
@@ -1,14 +1,21 @@
 [tox]
 envlist =
   py27
-  py34
+  py35
```
sakisv: Should we either add py34 here or remove it from the Dockerfile?

quis: Digital Marketplace (one of the few other users of the Python client) are running 3.6 in production. 

Let’s deal with this as a separate PR if we want to test against multiple versions. 3.5 is the best version to test against if we have to pick only one.

PR #105, opened 2018-04-20
Add note about last 7 days in get all

quis:
Follows: https://github.com/alphagov/notifications-ruby-client/pull/43


jonathanglassman:
@quis looks good to me.

quis:
Leaving this open as a reminder to do the other ones.

PR #111, opened 2018-05-11
Add support for attaching documents to email notifications

allait:
It's already possible to use the new Document Download integration
in the existing python client by setting a personalisation value to
`{"file": ...}`.

This simplifies the interface by recognizing file-like objects in
personalisation data and automatically reading/converting them to
the format expected by the Notify API:

```python

with open(<document-path>, 'rb') as f:
   client.send_email_notification(..., personalisation={'doc': f})

```

## Checklist

  - [x] `DOCUMENTATION.md`
  - [x] `CHANGELOG.md`
  - [ ] `CONTRIBUTING.md`

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -158,6 +158,17 @@ personalisation={
 }
 ```
 
+If your service has a permission to upload documents you can pass file objects as values for the personalisation keys:
+
+```python
+with open('file.pdf', 'rb') as f:
+    ...
+    personalisation={
+      'first_name': 'Amala',
+      'document': f,
+    }
+```
+
 #### reference (optional)
 
 A unique identifier. This reference can identify a single unique notification or a batch of multiple notifications.
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -181,26 +181,24 @@ email_reply_to_id='8e222534-7f05-4972-86e3-17c5d9f894e2' # optional UUID string
 
 If you omit this argument, the client will set your default email reply-to address for the notification.
 
-### Attach a document
+### Upload a document
 
-You can attach a document to an email notification.
+You can upload a document to link to from an email notification.
 
-Contact the GOV.UK Notify team on the support page or through the Slack channel to enable this function for your service.
+The document must be a PDF file smaller than 2MB.
 
-The document must be a PDF file smaller than 2MB. To attach the document, you must:
-
-1. Define the PDF as a file object.
-1. Pass the file object as a value into the personalisation argument. For example:
+To upload the document, pass the file object as a value into the personalisation argument. For example:
 
         ```python
-        with open('file.pdf', 'rb') as DOCUMENT_NAME:
+        with open('file.pdf', 'rb') as f:
             ...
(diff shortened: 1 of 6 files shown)


DOCUMENTATION.md
```diff
@@ -158,6 +158,17 @@ personalisation={
...
 ```
 
+If your service has a permission to upload documents you can pass file objects as values for the personalisation keys:
+
+```python
+with open('file.pdf', 'rb') as f:
+    ...
+    personalisation={
+      'first_name': 'Amala',
+      'document': f,
```
leohemsted: dont know if we want to document this process more at the client lib level to indicate that this magic file pointer thing will be a link in the output. probably not tho

quis: I think people will have questions, but doing the minimal thing for now lets us find out what those questions are in research 👍

quis:
Also in favour of the interface here. Means you can pass:
- `str`
- `list` (probably not documented?)
- or `io`

and Notify just does the right thing.

DOCUMENTATION.md
jonathanglassman: Had a chat with @richardc0 and @allait on this. We'll coordinate with Tom Chambers to make sure that the information about the feature is on the features page. I have a couple of questions about this and would like to answer those before this gets merged.

DOCUMENTATION.md
```diff
@@ -181,6 +181,27 @@ email_reply_to_id='8e222534-7f05-4972-86e3-17c5d9f894e2' # optional UUID string
...
 
+### Attach a document
+
+You can attach a document to an email notification.
+
+Contact the GOV.UK Notify team on the support page or through the Slack channel to enable this function for your service.
+
+The document must be a PDF file smaller than 2MB. To attach the document, you must:
+
+1. Define the PDF as a file object.
```
allait: I think this step is confusing and I'd argue it's not necessary. You don't define anything, you open an existing file, but that's not the only way to achieve this and it should be clear from the code sample.

DOCUMENTATION.md
```diff
@@ -181,6 +181,27 @@ email_reply_to_id='8e222534-7f05-4972-86e3-17c5d9f894e2' # optional UUID string
...
+1. Pass the file object as a value into the personalisation argument. For example:
+
+        ```python
+        with open('file.pdf', 'rb') as DOCUMENT_NAME:
```
allait: `f` is a conventional variable name for this. DOCUMENT_NAME is a bit misleading (since the value isn't actually a name) and it shouldn't be capitalize since it implies it's a global constant (which it isn't).

DOCUMENTATION.md
```diff
@@ -181,6 +181,27 @@ email_reply_to_id='8e222534-7f05-4972-86e3-17c5d9f894e2' # optional UUID string
 
 If you omit this argument, the client will set your default email reply-to address for the notification.
 
+### Attach a document
```
allait: Might be worth checking with @antimega about "attach" vs "upload". Email attachments are a specific thing and that's explicitly not what's going to happen in this case.

PR #115, opened 2018-07-23
Update responses when getting notifications and bump version 

klssmith:
The `get_notification_by_id` and `get_all_notifications` methods now
return a new field, `created_by_name`, so this has been documented and
the schema for the integration tests updated.

The version has been bumped from `4.9.0` to `4.10.0` as a result.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 4.10.0
+
+* Added `created_by_name` to the response for `NotificationsAPIClient.get_notification_by_id()` and `NotificationsAPIClient.get_all_notifications()`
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 4.10.0
+
+* Added `created_by_name` to the response for `NotificationsAPIClient.get_notification_by_id()` and `NotificationsAPIClient.get_all_notifications()`
+    * If the notification was sent manually, this will be the name of the sender. If the notification was uploaded as a CSV or was sent through the API, this will be `None`.
```
servingUpAces: I don't think we need to mention CSV uploads here because the jobs are not returned by these methods

PR #117, opened 2018-07-27
Change document upload API to use an explicit helper function

allait:
### Change document upload API to use an explicit helper function

The existing API requires the least amount of boilerplate, but we've seen a number of issues with it during user research:

* It's hiding the implementation details in a way that makes users suspect that there's some magic behind it and that 'document' key is special
* It requires the call to `post_notification` to be within the `with open(...)` block, which isn't immediately obvious, doesn't provide a very useful error message and makes it hard to create the personalisation dictionary beforehand.
* It relies on `isinstance` check which got a bit complex with python 2/3 file/io mismatches and might catch custom subclasses that users wouldn't expect to be treated as documents
* It might be hard to implement and would go against the conventions in statically-typed language clients like Java and C#.

The proposed helper function hopefully clarifies or solves some of these issues. The downside is that it requires a little more 'boilerplate' code and the function name has to be chosen carefully to not clash in the user's namespace and avoid giving the impression that it uploads the file (as opposed to transforming the value before it's sent to Notify API).

## Checklist

  - [x] `DOCUMENTATION.md`
  - [x] `CHANGELOG.md`
  - [ ] `CONTRIBUTING.md`


allait:
@quis while it would help better explain how the feature works overall, I think it suggests that the function returns a link, which it doesn't. So there's a risk of people expecting it to give them a link they could use or to upload the file before they send the notification.

I'm thinking of renaming the `'document'` key to something like `'link_placeholder'` in a separate PR to maybe address this and help remove the perception that 'document' key is special and doesn't need to be included in the template.

leohemsted:
need to bump vers num

antimega:
Looks good to me. ✅

PR #120, opened 2018-08-15
Updated email error codes

karlchillmaid:
Removed reference to 'attachments' from email error codes

Ticket: https://www.pivotaltracker.com/story/show/159434372

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -237,8 +237,8 @@ If the request is not successful, the client will return an `HTTPError` containi
 |:---|:---|:---|
 |`400`|`[{`<br>`"error": "BadRequestError",`<br>`"message": "Can't send to this recipient using a team-only API key"`<br>`]}`|Use the correct type of [API key](#api-keys)|
 |`400`|`[{`<br>`"error": "BadRequestError",`<br>`"message": "Can't send to this recipient when service is in trial mode - see https://www.notifications.service.gov.uk/trial-mode"`<br>`}]`|Your service cannot send this notification in [trial mode](https://www.notifications.service.gov.uk/features/using-notify#trial-mode)|
-|`400`|`[{`<br>`"error": "BadRequestError",`<br>`"message": "Unsupported document type '{}'. Supported types are: {}"`<br>`}]`|The attached document must be a PDF file|
-|`400`|`[{`<br>`"error": "BadRequestError",`<br>`"message": "Document didn't pass the virus scan"`<br>`}]`|The attached document must be virus free|
+|`400`|`[{`<br>`"error": "BadRequestError",`<br>`"message": "Unsupported document type '{}'. Supported types are: {}"`<br>`}]`|The document you upload must be a PDF file|
+|`400`|`[{`<br>`"error": "BadRequestError",`<br>`"message": "Document didn't pass the virus scan"`<br>`}]`|The document you upload must be virus free|
 |`403`|`[{`<br>`"error": "AuthError",`<br>`"message": "Error: Your system clock must be accurate to within 30 seconds"`<br>`}]`|Check your system clock|
 |`403`|`[{`<br>`"error": "AuthError",`<br>`"message": "Invalid token: signature, api token not found"`<br>`}]`|Use the correct type of [API key](#api-keys)|
 |`429`|`[{`<br>`"error": "RateLimitError",`<br>`"message": "Exceeded rate limit for key type TEAM/TEST/LIVE of 3000 requests per 60 seconds"`<br>`}]`|Refer to [API rate limits](#api-rate-limits) for more information|


DOCUMENTATION.md
```diff
@@ -185,13 +185,17 @@ email_reply_to_id='8e222534-7f05-4972-86e3-17c5d9f894e2' # optional UUID string
...
+### Send a document by email
 
-You can upload a document to link to from an email notification.
+You can send files without the need for email attachments. First upload a document, then send recipients a link to download it.
 
-The document must be a PDF file smaller than 2MB.
+To enable this function for your service [contact the GOV.UK Notify team](https://www.notifications.service.gov.uk/support).
 
-To upload the document, pass the file object as a value into the personalisation argument. For example:
+####Upload your document
```
karlchillmaid: `Upload *a* document` might be better

DOCUMENTATION.md
```diff
@@ -185,13 +185,17 @@ email_reply_to_id='8e222534-7f05-4972-86e3-17c5d9f894e2' # optional UUID string
 
 If you omit this argument, the client will set your default email reply-to address for the notification.
 
-### Upload a document
+### Send a document by email
 
-You can upload a document to link to from an email notification.
+You can send files without the need for email attachments. First upload a document, then send recipients a link to download it.
```
jonathanglassman: We could make it more positive - "You can send a document by sending an email with a link to that document".

DOCUMENTATION.md
```diff
@@ -185,13 +185,17 @@ email_reply_to_id='8e222534-7f05-4972-86e3-17c5d9f894e2' # optional UUID string
...
 
-The document must be a PDF file smaller than 2MB.
+To enable this function for your service [contact the GOV.UK Notify team](https://www.notifications.service.gov.uk/support).
```
jonathanglassman: Perhaps "[Contact the GOV.UK Notify team] to enable this function for your service." Puts the action up front.

DOCUMENTATION.md
jonathanglassman: Agree

DOCUMENTATION.md
```diff
@@ -205,7 +209,9 @@ with open('file.pdf', 'rb') as f:
     }
 ```
 
-Contact the GOV.UK Notify team on the support page or through the Slack channel to enable this function for your service.
+####Add a link to the email template
+
+To include a link to the document in your email notification, add a placeholder to the template in Notify.
```
jonathanglassman: What sort of placeholder? Is there any particular syntax that must be used?

DOCUMENTATION.md
```diff
@@ -339,7 +345,7 @@ If the request is not successful, the client will return an `HTTPError` containi
 |:---|:---|:---|
 |`400`|`[{`<br>`"error": "BadRequestError",`<br>`"message": "Cannot send letters with a team api key"`<br>`]}`|Use the correct type of [API key](#api-keys)|
 |`400`|`[{`<br>`"error": "BadRequestError",`<br>`"message": "Cannot send letters when service is in trial mode - see https://www.notifications.service.gov.uk/trial-mode"`<br>`}]`|Your service cannot send this notification in  [trial mode](https://www.notifications.service.gov.uk/features/using-notify#trial-mode)|
-|`400`|`[{`<br>`"error": "ValidationError",`<br>`"message": "personalisation address_line_1 is a required property"`<br>`}]`|Ensure that your template has a field for the first line of the address, check [personlisation](#send-a-letter-arguments-personalisation-optional) for more information.|
+|`400`|`[{`<br>`"error": "ValidationError",`<br>`"message": "personalisation address_line_1 is a required property"`<br>`}]`|Ensure that your template has a field for the first line of the address, check [personalisation](#send-a-letter-arguments-personalisation-optional) for more information.|
```
jonathanglassman: Is this error code missing from anywhere else in this document or others?

DOCUMENTATION.md
karlchillmaid: Agree that it could be more positive. However we felt there was a need to mention email attachments – and explain that this feature is different – because many users will expect attachments to be the solution to sending docs.

Also agree that we could be clearer about the fact that this adds a link to the doc to an email.

DOCUMENTATION.md
karlchillmaid: It's the ((double brackets)) placeholder used for personalisation in the template.
There will be formatting guidance on the Edit template screen, but I'd like to include an example here too. Need to confirm exactly what the placeholder should be first.

DOCUMENTATION.md
karlchillmaid: Not sure. I was just correcting a typo in the link text.

DOCUMENTATION.md
```diff
@@ -185,13 +185,19 @@ email_reply_to_id='8e222534-7f05-4972-86e3-17c5d9f894e2' # optional UUID string
...
-You can upload a document to link to from an email notification.
+To send a document by email, upload a file then add a placeholder field to the template.
 
-The document must be a PDF file smaller than 2MB.
+The placeholder field contains a secure link to download the document instead of an attachment.
 
-To upload the document, pass the file object as a value into the personalisation argument. For example:
+[Contact the GOV.UK Notify team](https://www.notifications.service.gov.uk/support) to enable this function for your service.
+
+####Upload your document
```
allait: This needs a space after `####` to render as a header

DOCUMENTATION.md
```diff
@@ -205,7 +211,9 @@ with open('file.pdf', 'rb') as f:
     }
 ```
 
-Contact the GOV.UK Notify team on the support page or through the Slack channel to enable this function for your service.
+####Add a link to the email template
```
allait: Space after `####` to render as a header

DOCUMENTATION.md
```diff
@@ -205,7 +211,9 @@ with open('file.pdf', 'rb') as f:
...
+
+To include a link to the document in your email notification, add a placeholder field to the template in Notify.
```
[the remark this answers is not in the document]
allait: It's `link_to_document` in the code example we give https://github.com/alphagov/notifications-python-client/pull/120/files#diff-dc89fdcc242c31d756fdf63cc2b6a434R210. But it has to be clear it's just an example or we'll be back to users thinking the key is somehow special.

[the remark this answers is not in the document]
karlchillmaid: So –just to make sure I've understood properly– in the example given, users would need to add `((link_to_document))` to the template in Notify?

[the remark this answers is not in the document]
allait: Yes. It will get replaced with a link so hopefully the template has some context for it around the placeholder.

DOCUMENTATION.md
```diff
@@ -191,15 +191,21 @@ For example:
...
+You can leave out this argument if your service only has one email reply-to address, or you want to use the default email address.
 
-### Upload a document
+### Send a document by email
 
-You can upload a document to link to from an email notification.
+To send a document by email, upload a file then add a placeholder field to the template.
 
-The document must be a PDF file smaller than 2MB.
+The placeholder field contains a secure link to download the document instead of an attachment.
```
allait: "will contain" or "will be replaced with a link"? I feel like this sentence is trying to say more than one thing too: what happens with the placeholder value and that it's different from attachments.

DOCUMENTATION.md
```diff
@@ -213,7 +219,11 @@ with open('file.pdf', 'rb') as f:
     }
 ```
 
-Contact the GOV.UK Notify team through the [support page](https://www.notifications.service.gov.uk/support) or the [Slack channel](https://ukgovernmentdigital.slack.com/messages/govuk-notify) to enable this function for your service.
+#### Add a placeholder field to the email template
```
allait: I think this section should go before 'Upload your document'

DOCUMENTATION.md
```diff
@@ -191,15 +191,21 @@ For example:
...
 ```
 
-You can leave out this argument if your service only has one email reply to address, or you want to use the default email address.
+You can leave out this argument if your service only has one email reply-to address, or you want to use the default email address.
 
-### Upload a document
+### Send a document by email
 
-You can upload a document to link to from an email notification.
+To send a document by email, upload a file then add a placeholder field to the template.
```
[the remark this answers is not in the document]
karlchillmaid: Thanks for clarifying – I'll swap that round.

DOCUMENTATION.md
karlchillmaid: I agree that the sentence is trying to do too many things.

User research showed that we need to explain that this meets the need for email attachments, but is not _actually_ email attachments.

Once we have some feature content for _send a document_ on Notify, I'd like to include a link to that here. Eg. "Find out why we use links instead of attachments." However, until that's ready we need an alternative solution.

An earlier iteration of the _send a document_ section opened with the line `Send files without the need for email attachments. 
To send a document by email...` 

We could revert to that and then change this line to `The placeholder field is replaced by a secure link to download the document.`

What do you think @allait @jonathanglassman?

PR #121, opened 2018-08-23
Add missing bold formatting

jonathanglassman:

Adds missing bold formatting to the Python > Send a letter docs

## Checklist

- [ ] I’ve updated the documentation in
  - [x] `DOCUMENTATION.md`
  - [ ] `CHANGELOG.md`
  - [ ] `CONTRIBUTING.md`


quis:
There’s some repetition going on here:

image

***

Also I don’t think this section is as accurate/helpful as it could be. You can’t send real letters in trial mode, but you can use a test key to fake sending letters, and see what they would look like.

So we should emphasise that, not what the process for going live is.

jonathanglassman:
This needs to be amended for all clients then.

quis:
Merging this for the **bold** changes. Can revisit the letters stuff.

PR #122, opened 2018-09-03
Remove empty payload

sakisv:

Don't send an empty payload

Currently all our get/post/put/delete methods are an abstraction over a
generic `request` method which defaults to `None` for its `data` and
`params` arguments.

While each of the get/post/put/delete use only one of these arguments,
that has the uninteneded side-effect that GET requests are sent with a
body which is the json representation of `None`:

```python
>>> import json
>>> json.dumps(None)
'null'
```

This behaviour causes problems with Cloudfront which does not accept GET
requests that include a body and returns a 403 [1].

1: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/RequestAndResponseBehaviorCustomOrigin.html#RequestCustom-get-body

## Checklist

  - [ ] ~`DOCUMENTATION.md`~
  - [x] `CHANGELOG.md`
- [ ] ~I've added new environment variables to~
  - [ ] ~`notifications-python-client/scripts/generate_docker_env.sh`~
  - [ ] ~`notifications-python-client/tox.ini`~
  - [ ] ~`CONTRIBUTING.md`~

### DIFF
--- a/notifications_python_client/base.py
+++ b/notifications_python_client/base.py
@@ -65,12 +65,20 @@ class BaseAPIClient(object):
 
         logger.debug("API request {} {}".format(method, url))
 
-        payload = json.dumps(data)
-
         api_token = create_jwt_token(
             self.api_key,
             self.service_id
         )
+
+        kwargs = {
+            "headers": self.generate_headers(api_token),
+        }
+
+        if data is not None:
+            kwargs.update(data=json.dumps(data))
+
+        if params is not None:
+            kwargs.update(params=params)
 
         url = urllib.parse.urljoin(str(self.base_url), str(url))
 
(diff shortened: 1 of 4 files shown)


notifications_python_client/base.py
```diff
@@ -65,23 +65,29 @@ def request(self, method, url, data=None, params=None):
...
         api_token = create_jwt_token(
             self.api_key,
             self.service_id
         )
 
+        kwargs = {
+            "headers": self.generate_headers(api_token),
+        }
+
+        if data:
```
allait: I'd go for `is not None` for these checks. We're unlikely to ever use `0` or `False` as the payload, but you never know.

sakisv: Good point, done and rebased 👍

PR #126, opened 2018-09-12
Add prepare upload file size check

allait:
### Check that uploaded documents are smaller than the size limit

Since we can't guarantee that users will see the original 413 error message from document download we're implementing a file size check in the `prepare_upload` helper in all API clients.

### Bump version to 5.2.0

This is technically a breaking change since it's replacing one exception with another, but we know `prepare_upload` is unused at the moment and after `5.0.1` we need to make it less scary for users to upgrade clients.

## Checklist

  - [ ] `DOCUMENTATION.md`
  - [x] `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 5.2.0
+
+* Add a document size check to `prepare_upload`. Will raise `ValueError` when trying to upload a document larger than 2MB.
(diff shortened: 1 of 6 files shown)


klssmith:
The new error message could be added to the errors table in the documentation file.

PR #135, opened 2018-12-10
Updated message status

jonathanglassman:

Updated message statuses to align with the "Using Notify" page and the Support docs.

## Checklist

- [x] I’ve updated the documentation (in `DOCUMENTATION.md`)
- [ ] I’ve bumped the version number (in `src/main/resources/application.properties`)
      and run the `update_version.sh` script


DOCUMENTATION.md
```diff
@@ -439,32 +439,32 @@ You can only get the status of messages that are 7 days old or newer.
...
+|Failed|This covers all failure statuses:<br>- `permanent-failure` - "The provider could not deliver the message because the email address or phone number was wrong. You should remove these email addresses or phone numbers from your database. You’ll still be charged for text messages to numbers that do not exist."<br>- `temporary-failure` - "The provider could not deliver the message. This can happen when the recipient's inbox is full, or their phone is off. You can try to send the message again."<br>- `technical-failure` - "Your message was not sent because there was a problem between Notify and the provider.<br>You’ll have to try sending your messages again. You will not be charged for text messages that are affected by a technical failure."|
```
karlchillmaid: Have we dropped the bit about trying for 72 hours on 'temporary failure' to save space?

DOCUMENTATION.md
```diff
@@ -439,32 +439,32 @@ You can only get the status of messages that are 7 days old or newer.
...
+|Sent|The message was sent to an international number. The mobile networks in some countries do not provide any more delivery information.|
 
 ## Status - letter
 
 |Status|information|
 |:---|:---|
 |Failed|The only failure status that applies to letters is `technical-failure`. GOV.UK Notify had an unexpected error while sending to our printing provider.|
-|Accepted|GOV.UK Notify is printing and posting the letter.|
-|Received|The provider has received the letter to deliver.|
+|Accepted|GOV.UK Notify has placed the letter in a queue, ready to be sent to the provider.|
```
karlchillmaid: We need to iterate this.

DOCUMENTATION.md
```diff
@@ -439,32 +439,32 @@ You can only get the status of messages that are 7 days old or newer.
...
+|Received|GOV.UK Notify has sent the letter to the provider. The provider has printed and dispatched the letter.|
```
karlchillmaid: We need to iterate this – it's not entirely accurate.

DOCUMENTATION.md
```diff
@@ -439,32 +439,32 @@ You can only get the status of messages that are 7 days old or newer.
...
 
 ## Status - precompiled letter
 
 |Status|information|
 |:---|:---|
-|Pending virus check|GOV.UK Notify virus scan of the precompiled letter file is not yet complete.|
-|Virus scan failed|GOV.UK Notify virus scan has identified a potential virus in the precompiled letter file.|
+|Pending virus check|GOV.UK Notify has not completed a virus scan of the pre-compiled letter file.|
```
karlchillmaid: I have a question for our developers about lines 466 and 467.
The descriptions are fine (I think), but I'd like to know if we can make the statuses consistent, so we don't say virus check *and* virus scan.

DOCUMENTATION.md
```diff
@@ -439,32 +439,32 @@ You can only get the status of messages that are 7 days old or newer.
 
 |Status|Information|
 |:---|:---|
-|Created|The message is queued to be sent to the provider. The notification usually remains in this state for a few seconds.|
-|Sending|The message is queued to be sent by the provider to the recipient, and GOV.UK Notify is waiting for delivery information.|
+|Created|GOV.UK Notify has placed the message in a queue, ready to be sent to the provider. It should only remain in this state for a few seconds.|
+|Sending|GOV.UK Notify has sent the message to the provider. The provider will try to deliver the message to the recipient. GOV.UK Notify is waiting for delivery information.|
```
karlchillmaid: @jonathanglassman For now this status is still `sending` – I'm trying to find out how soon we can change it to `Sent` as discussed in our last meeting.

jonathanglassman:
@karlchillmaid I agree. Once @servingUpAces reviews and we agree on final text, I can duplicate to other clients.

DOCUMENTATION.md
```diff
@@ -439,32 +439,32 @@ You can only get the status of messages that are 7 days old or newer.
...
 
 ## Status - text only
 
 |Status|Information|
 |:---|:---|
-|Pending|GOV.UK Notify received a callback from the provider but the device has not yet responded. Another callback from the provider determines the final status of the notification.|
-|Sent|The text message was delivered internationally. This only applies to text messages sent to non-UK phone numbers. GOV.UK Notify may not receive additional status updates depending on the recipient's country and telecoms provider.|
+|Pending|GOV.UK Notify is waiting for more delivery information.<br>GOV.UK Notify received a callback from the provider but the recipient's device has not yet responded. Another callback from the provider determines the final status of the notification.|
+|Sent internationally|The message was sent to an international number. The mobile networks in some countries do not provide any more delivery information.|
```
servingUpAces: The API will return this status as `sent`

karlchillmaid:
@servingUpAces 
> There is another status called `validation-failed`. This is when the precompiled PDF fails our validation, usually there is something in the margin that needs to be left blank.

I think I wrote the following confirmation and error messages for that:

`Your letter passed the layout check`
`Content in this PDF is outside the printable area`

Is that what this status refers to?

DOCUMENTATION.md
karlchillmaid: screen shot 2018-12-20 at 09 28 59
Seems like it does.

@jonathanglassman do we want to say something slightly different in the tech docs and app to cover this?

jonathanglassman: Could put in both, as a dev using the tech docs might use both the API and the app and see different things in both

jonathanglassman: Amended. @servingUpAces, @karlchillmaid, please take a look

jonathanglassman:
Added a validation failed status, please review when you have a chance

karlchillmaid:
@jonathanglassman 

Can we be more specific? For example:

>Content in the precompiled letter file is outside the printable area.

jonathanglassman:
@karlchillmaid looks good to me - sorry, didn't see your comment

DOCUMENTATION.md
```diff
@@ -439,32 +439,33 @@ You can only get the status of messages that are 7 days old or newer.
...
+|Accepted|GOV.UK Notify has sent the letter to the provider to be printed.|
+|Received|The provider has printed and dispatched the letter.|
 
 ## Status - precompiled letter
 
 |Status|information|
 |:---|:---|
-|Pending virus check|GOV.UK Notify virus scan of the precompiled letter file is not yet complete.|
-|Virus scan failed|GOV.UK Notify virus scan has identified a potential virus in the precompiled letter file.|
+|Pending virus check|GOV.UK Notify has not completed a virus scan of the precompiled letter file.|
```
servingUpAces: should be `pending-virus-check`, `virus-scan-failed` and `validation-failed`

jonathanglassman: Do we need to phrase it as it would look in the API? I thought these tables were non-code formatted as they are descriptions rather than code blocks.

karlchillmaid: I thought these tables were non-code formatted too – because they provide information about statuses, rather than instructions/method/arguments.

The statuses are formatted as code in the _Get the status of multiple messages_ section: https://docs.notifications.service.gov.uk/python.html#get-the-status-of-multiple-messages

jonathanglassman: Becca confirmed PR is fine on slack at 1056am on 23/01/2019

PR #156, opened 2020-01-07
Catch generic JWT InvalidTokenErrors

idavidmcdonald:
- We are seeing niche tokens come in (for example users setting the `aud` claim) which fail validation by jwt and raise exceptions that we aren't explicitely catching. We want to catch these exceptions and so therefore we catch all `InvalidTokenError`s to cover invalid tokens that aren't being caught by our more specific errors.

## Questions
- Should all our errors to do with tokens also provide the link to the docs (on top of their more specific error message) as this seems generally useful?

## Checklist

  - [ ] `DOCUMENTATION.md`
  - [x] `CHANGELOG.md`
  - [ ] `CONTRIBUTING.md`

### DIFF
--- a/notifications_python_client/errors.py
+++ b/notifications_python_client/errors.py
@@ -16,13 +16,18 @@ class TokenError(Exception):
         self.token = token
 
 
+class TokenExpiredError(TokenError):
+    pass
+
+
+class TokenAlgorithmError(TokenError):
+    def __init__(self):
+        super().__init__('Invalid token: algorithm used is not HS256')
+
+
 class TokenDecodeError(TokenError):
     def __init__(self, message=None):
         super().__init__(message or 'Invalid token: signature')
-
-
-class TokenExpiredError(TokenError):
-    pass
 
 
 class TokenIssuerError(TokenDecodeError):
@@ -33,11 +38,6 @@ class TokenIssuerError(TokenDecodeError):
 class TokenIssuedAtError(TokenDecodeError):
     def __init__(self):
         super().__init__('Invalid token: iat field not provided')
-
-
-class TokenAlgorithmError(TokenDecodeError):
-    def __init__(self):
-        super().__init__('Invalid token: algorithm used is not HS256')
 
 
 class APIError(Exception):
--- a/notifications_python_client/errors.py
+++ b/notifications_python_client/errors.py
@@ -5,14 +5,17 @@ from __future__ import absolute_import
 from builtins import super
 from future import standard_library
 standard_library.install_aliases()
+
 REQUEST_ERROR_STATUS_CODE = 503
 REQUEST_ERROR_MESSAGE = "Request failed"
+
+TOKEN_ERROR_GENERIC_MESSAGE = "Invalid token: See our requirements for JSON Web Tokens at https://docs.notifications.service.gov.uk/rest-api.html#authorisation-header"  # noqa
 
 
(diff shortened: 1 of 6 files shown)


notifications_python_client/errors.py
```diff
@@ -5,26 +5,34 @@
 from builtins import super
 from future import standard_library
 standard_library.install_aliases()
+
 REQUEST_ERROR_STATUS_CODE = 503
 REQUEST_ERROR_MESSAGE = "Request failed"
 
+TOKEN_ERROR_GENERIC_MESSAGE = "Invalid token: see https://docs.notifications.service.gov.uk/rest-api.html#authorisation-header"  # noqa
```
CrystalPea: Maybe the message could be something like: 
"Invalid token. For token specifications see https://docs.notifications.service.gov.uk/rest-api.html#authorisation-header" ?

@karlchillmaid what do you think?

tests/notifications_python_client/test_authentication.py
```diff
@@ -49,6 +56,26 @@ def test_token_should_fail_to_decode_if_wrong_key():
     assert str(err.value) == "Signature verification failed"
 
 
+@pytest.mark.parametrize('exc_class', [
```
CrystalPea: what's `exc`? Is there a more accessible word for it?

notifications_python_client/errors.py
idavidmcdonald: Oh yes, good point. I'd almost forgotten to say I'd definitely like your review on the content please @karlchillmaid

tests/notifications_python_client/test_authentication.py
idavidmcdonald: Good point, will make this more readable

notifications_python_client/errors.py
karlchillmaid: Would this be too long?

`Invalid token. Find out how to create the authorisation header: https://docs.notifications.service.gov.uk/rest-api.html#authorisation-header`

I'm just trying to get a bit of continuity between the link text and the page/heading users land on.

idavidmcdonald: I think length is fine but I might suggest something in the middle ground:

`Invalid token. Find out how to create a valid JSON Web token: https://docs.notifications.service.gov.uk/rest-api.html#authorisation-header`

As the issue isn't that they aren't sending an authorisation header, it's that the token which is the value of that header is not a valid JWT token.

karlchillmaid: @idavidmcdonald Yeah, that's a more helpful message. It specifically addresses the problem they're trying to solve. Given the fact we don't want the message to get too long, we can forget about flagging the 'Authorisation header' heading in advance.

idavidmcdonald: Sorry, I didn't quite understand your last sentence. Are you saying that we should drop `#authorisation-header` from the provided link?

klssmith:
Not sure if it's worth having an entry for this version in the Changelog even though this change doesn't affect any teams using this client. Just so that if they're prompted to upgrade, e.g. by Pyup or Dependabot, they can see what has changed easily

idavidmcdonald:
@karlchillmaid This is ready for a second review of the content please

@klssmith I've also improved other error messages so I'm going to bump it a minor version and have added a changelog record

@CrystalPea ready for a final technical review please

PR #160, opened 2020-02-24
Make send files by email self service

karlchillmaid:

We need to update our documentation to reflect work done for the [Allow users to turn on 'Send files by email'](https://www.pivotaltracker.com/story/show/171254127) story.

This PR:
- adds instructions for how to add contact details to the file download page in Settings
- removes the old 400 error `Service is not allowed to send documents`
- adds a new 400 error, triggered if the service has not added contact details yet
- fixes some broken links in the error message tables

## Checklist

  - [ ] `DOCUMENTATION.md`
  - [ ] `CHANGELOG.md`
  - [ ] `CONTRIBUTING.md`


DOCUMENTATION.md
```diff
@@ -206,20 +206,23 @@ You can leave out this argument if your service only has one reply-to email addr
...
 ## Send a file by email
 
-Send files without the need for email attachments.
+To send a file by email, add a placeholder to the template then upload a file. The placeholder will contain a secure link to download the file.
 
-This is an invitation-only feature. [Contact the GOV.UK Notify team](https://www.notifications.service.gov.uk/support/ask-question-give-feedback) to enable this function for your service.
+The links are unique and unguessable. GOV.UK Notify cannot access or decrypt your file.
 
-To send a file by email, add a placeholder field to the template then upload a file. The placeholder field will contain a secure link to download the file.
+#### Add contact details to the file download page
```
CrystalPea: under this header, should we maybe add one sentence explaining that they need to do it to use the feature? Something like: "To send files by email, you first need to add contact details to the file download page, so the users can get in touch if there is a problem with the file."

karlchillmaid: We could reuse the content users will see on the _Send files by email_ settings page?

>You need to include contact details for your service so your users can get in touch if there’s a problem. For example, if the link to download the file you sent them has expired.

The reason we didn't include something like this already was to keep things simple by sticking to the **how** rather than the **why**. 

We can add this content if we think it's necessary, but is it too much detail for the documentation, especially if users will see the explanation as soon as they go to their settings?

Put another way: do we think users question the instructions we give them in the documentation? And, if they do, would it stop them following the instructions?

CrystalPea: I see,  I think what I meant is more to indicate that it's mandatory, because it does not sound mandatory, but maybe it's just me.

karlchillmaid: I think that's my question, really: do we have to **say** something's mandatory for users to do it?

In the arguments section we state which are `(required)` and which are `(optional)` – do we need to start doing something similar for general instructions?

CrystalPea: Hmm I think it depends if there is a mental leap or no. Like sending files by email does not seem directly connected to setting up contact details. But I might be wrong.

karlchillmaid: My gut is to leave it, and see if anyone has trouble with it.
If they do ignore/forget to do this, they'll trigger the error message, which will get them back on track.
Is there a way for us to measure if/when that error is triggered by the API?

CrystalPea: Ok, let's do it :). Yeah, we can check that in our logs for last two weeks :)

Issue #164, opened 2020-04-23
A proposal for fixing CSV downloads in notify


currycoder:
**The Problem**

GOVUK notify (https://www.notifications.service.gov.uk/) is a web service we are using to send emails to external partners. For our use case, we need to send these partners spreadsheet files (CSVs) in an agreed format for them to process and send back to us. GOVUK notify offers a mechanism for file uploads/downloads which we would like to use for this purpose.

Unfortunately, CSV files uploaded as part of a notify email message end up being saved as TXT files. This means that our partners download these files as TXT files, which is against our agreed process and is causing us problems.

The problematic bit of code lies in GOVUK's document-download-api web service, here: https://github.com/alphagov/document-download-api/blob/master/app/upload/views.py#L18

The `get_mime_type()` utility acts based on the contents of a file only (at this point in the flow, the service does not have access to file name/extension) - and so it identifies the file as a TXT file instead of as a CSV.

**Proposed Solution**

We need to provide a solution here which works both for our particular situation and for existing notify users (who may have worked around this problem in their own ways).

The proposal is for us to adjust `notifications-python-client`, `notifications-api` and `document-download-api` so that we can optionally pass the correct file mimetype through to the logic that saves the file.

This would require the following changes:

- The `prepare_upload()` utility in `notifications-python-client` would need to optionally take a mimetype kwarg: https://github.com/alphagov/notifications-python-client/blob/ddf6b1afb4b8bb22badc835a93c4f95a33eb2ccf/notifications_python_client/utils.py#L6. This should result in the `mimetype` being added to the return dictionary e.g. `{'data': '<base64>', 'mimetype': 'text/csv'}`.
- The API endpoint for sending email notifications on `notifications-api` would need adjusting to accept mimetype being passed through in JSON objects representing files: https://github.com/alphagov/notifications-api/blob/master/app/v2/notifications/post_notifications.py#L117
- The method that `notifications-api` uses to upload documents to `document-download-api` would need adjusting to accept and forward mimetype as part of the POST body. https://github.com/alphagov/notifications-api/blob/d88b20beeca799f920eb84fb1695f336e4fb5a29/app/clients/document_download.py#L29
[...]

PR #165, opened 2020-04-30
Add is_csv kwarg to prepare_upload util

currycoder:

This PR adds an optional `is_csv` parameter to the `prepare_upload()` utility to allow clients to explicitly specify that the file they are uploading is a CSV. This provides a fix for an issue where uploads could not be saved as CSV and would always be saved as TXT. For more info, refer to: alphagov/notifications-python-client#164

## Checklist

- [X] I’ve written unit tests for these changes
- [X] I’ve update the documentation in
  - [X] `DOCUMENTATION.md`
  - [X] `CHANGELOG.md`
- [X] I’ve bumped the version number in
  - [X] `notifications_python_client/__init__.py`
  - [X] `notifications_python_client/test_base_api_client.py`, function `test_user_agent_is_set`
  - [ ] `CONTRIBUTING.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 5.6.0
+
+* Add support for an optional `is_csv` parameter in the `prepare_upload()` function. This fixes a bug when sending a CSV file by email. This ensures that the file is downloaded as a CSV rather than a TXT file.
+
 ## 5.5.1
 
 * change error message to refer to file rather than document
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -244,6 +244,23 @@ with open('file.pdf', 'rb') as f:
     }
 ```
 
+#### CSV Files
+
+Uploads for CSV files should use the `is_csv` parameter
+on the `prepare_upload()` utility.  For example:
+
+```python
+from notifications_python_client import prepare_upload
+
+with open('file.csv', 'rb') as f:
+    ...
+    personalisation={
+      'first_name': 'Amala',
+      'application_date': '2018-01-01',
+      'link_to_file': prepare_upload(f, is_csv=True),
+    }
+```
+
 ### Response
 
 If the request to the client is successful, the client returns a `dict`:
--- a/notifications_python_client/__init__.py
+++ b/notifications_python_client/__init__.py
@@ -15,7 +15,7 @@ standard_library.install_aliases()
 #
 # -- http://semver.org/
 
-__version__ = '5.5.1'
+__version__ = '5.6.0'
 
 from notifications_python_client.errors import REQUEST_ERROR_STATUS_CODE, REQUEST_ERROR_MESSAGE  # noqa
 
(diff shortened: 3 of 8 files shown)


currycoder:
This PR is blocked waiting on: https://github.com/alphagov/document-download-api/pull/129

The PR for document-download-api will give users a better experience when downloading CSVs from notifications.

currycoder:
This is now unblocked.

DOCUMENTATION.md
```diff
@@ -244,6 +244,23 @@ with open('file.pdf', 'rb') as f:
     }
 ```
 
+#### CSV Files
+
+Uploads for CSV files should be handled a little differently - using the `is_csv` parameter
```
karlchillmaid, line 249:
```suggestion
Uploads for CSV files should use the `is_csv` parameter
```

currycoder:
@karlchillmaid Thanks for the comment!  Do you have any other thoughts on the docs?

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 5.6.0
+
+* Add support for optional parameter `is_csv` in `prepare_upload()` file upload utility function. This fixes a bug to ensure that the file is downloaded as a CSV rather than a TXT file.
```
karlchillmaid, line 3:
```suggestion
* Fix a bug in Send a file by email. Add support for an optional `is_csv` parameter in the `prepare_upload()` utility. This ensures that the file is downloaded as a CSV rather than a TXT file.
```

currycoder: Looks good to me!

idavidmcdonald: Thanks Karl. 

Wonder if it should be 'in the `prepare_upload()` function' or just 'in `prepare_upload()`' rather than 'in the `prepare_upload()` utility' though? I don't think the word utility is meaningful

currycoder: That does make better sense to me - I've updated with your initial suggestion there.  Does that sound OK?

karlchillmaid: That's where your subject matter expertise comes in very handy @idavidmcdonald.

'Utility' isn't meaningful to me either, but I wasn't sure if it was a specific term.

Let's go with your suggestion.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 5.6.0
+
+* Fix a bug in Send a file by email. Add support for an optional `is_csv` parameter in the `prepare_upload()` function. This ensures that the file is downloaded as a CSV rather than a TXT file.
```
idavidmcdonald, line 3: Final edit. Just spoke with Karl and we think the sentences are too disjoint so we now suggest:

Add support for an optional is_csv parameter in the prepare_upload() function. This fixes a bug when sending a CSV file by email. This ensures that the file is downloaded as a CSV rather than a TXT file.

After that then this is definitely going to be good to go. Thanks for your patience!

PR #172, opened 2020-07-15
Fix parsing of errors from Notify REST API

lfdebrux:

Currently the notifications client HTTPError exception will return the `errors` object for errors that look like those documented in https://docs.notifications.service.gov.uk/rest-api.html#error-codes.

This isn't what I was expecting, and seems like a bug to me, for two reasons:

1. the `str()` value will include the JSON in a way that looks ugly
2. the type of `error` will change depending on the format of the HTTP
response (it can be a string or a list of dicts)

This commit tries to fix this in a backwards-compatible way, by adding two new properties, `errors` that always returns a list of dicts, and `error_message` that always returns a `str`. `str()` of a HTTPError will now always just include the error message, rather than the whole error JSON object.

## Checklist

  - [ ] `DOCUMENTATION.md`
  - [ ] `CHANGELOG.md`
  - [ ] `CONTRIBUTING.md`

### DIFF
--- a/notifications_python_client/__init__.py
+++ b/notifications_python_client/__init__.py
@@ -15,7 +15,7 @@ standard_library.install_aliases()
 #
 # -- http://semver.org/
 
-__version__ = '5.7.0'
+__version__ = '5.8.0'
 
 from notifications_python_client.errors import REQUEST_ERROR_STATUS_CODE, REQUEST_ERROR_MESSAGE  # noqa
 
--- a/notifications_python_client/errors.py
+++ b/notifications_python_client/errors.py
@@ -50,7 +50,21 @@ class APIError(Exception):
         self._message = message
 
     def __str__(self):
-        return "{} - {}".format(self.status_code, self.message)
+        return "{} - {}".format(self.status_code, self.error_message)
+
+    @property
+    def errors(self):
+        try:
+            return self.response.json().get('errors', [self.response.json()])
+        except (TypeError, ValueError, AttributeError, KeyError):
+            return [{'message': self._message or REQUEST_ERROR_MESSAGE}]
+
+    @property
+    def error_message(self):
+        try:
+            return self.response.json()['errors'][0]['message']
+        except (TypeError, ValueError, AttributeError, IndexError, KeyError):
+            return self.message
 
     @property
     def message(self):
--- a/tests/notifications_python_client/test_base_api_client.py
+++ b/tests/notifications_python_client/test_base_api_client.py
@@ -82,6 +82,31 @@ def test_non_2xx_response_raises_api_error(base_client, rmock):
     assert e.value.status_code == 404
 
 
+def test_exception_parses_errors_response(rmock, base_client):
+    """Test parsing of error response"""
+    rmock.request(
+        "GET",
+        "http://test-host",
+        json={
+            "errors": [{
+                "error": "AuthError",
+                "message": "Unauthorized: authentication token must be provided"
+            }],
+            "status_code": 401
+        },
+        status_code=401
+    )
+
+    with pytest.raises(HTTPError) as e:
+        base_client.request("GET", "/")
+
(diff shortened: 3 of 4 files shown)


benthorner:
This PR has been around for a while, do you want to get this reviewed?

lfdebrux:
@benthorner hi Ben, a review would if possible, even it it's just to say that this change is not the way you'd like to go.

benthorner:
Hmm...tricky. I had a look around and can see:

- Our API does indeed multiple error response formats (like you point out) [[1](https://github.com/alphagov/notifications-api/blob/f85dad5acfa7e6da4bb64aaa82e918c114b6dd5f/app/errors.py#L57)] [[2](https://github.com/alphagov/notifications-api/blob/f85dad5acfa7e6da4bb64aaa82e918c114b6dd5f/app/errors.py#L30)]. Neither are going to change any time soon.
- A [similar change](https://github.com/alphagov/notifications-ruby-client/pull/73/commits/52399fb11221b7944fe8a59b7145c49ef8bd3f67) was made in the Ruby client 2 years ago.

I'm wondering if, for consistency, we should copy the approach there, which is similar to yours, but copes with multiple errors. Given that change wasn't backwards compatible, I think we've got a precedent for a breaking change here as well. This means we shouldn't need the `error_message` property. 

One other quick thought: is there a specific scenario for the `or REQUEST_ERROR_MESSAGE` bit? I noticed there's no such default in the Ruby client, so perhaps we don't need it.

lfdebrux:
> I'm wondering if, for consistency, we should copy the approach there, which is similar to yours, but copes with multiple errors. Given that change wasn't backwards compatible, I think we've got a precedent for a breaking change here as well. This means we shouldn't need the error_message property.

Copying the Ruby API interface and making `message` a string makes a lot of sense to me.

> One other quick thought: is there a specific scenario for the or REQUEST_ERROR_MESSAGE bit? I noticed there's no such default in the Ruby client, so perhaps we don't need it.

I'm afraid I don't know the answer to that.

lfdebrux:
@benthorner if you want I can open a new PR which makes the changes you suggest

benthorner:
@lfdebrux that sounds good to me, thanks! It's not a change we were planning to make, but I think it's an improvement and I'm happy to review it. Just because I'm quite new to the team, I'd like someone else from the team to comment, just in case I've not thought of something (e.g. maybe we know some users who depend on the _exact_ error message).

@klssmith would you like to comment on my suggestion?

> I'm afraid I don't know the answer to that.

Ah, I didn't notice this is the default for the `message` property. Makes sense now 👍.

klssmith:
@benthorner I'm not sure if any users depend on the exact message - probably a good one to check on Slack in case others in the team know

leohemsted:
i've mentioned this internally in our team slack, but I don't believe that in our current API code we ever return failure json of the format `{"message": "Some wordy description", "result": "error"}`. The only exception (ho ho) to that might be if Notify is down and our nginx returns a prewritten error message, but we can (should) update that if that's the case.

benthorner:
@leohemsted thanks for your insight here. For everyone else, the previous comment is challenging my assertion "Our API does indeed multiple error response formats". I've had a look at this myself, and I agree it's not the case (anymore). Leo pointed out a minor edge case where we may still return a string error response ([here](https://github.com/alphagov/notifications-aws/blob/master/ansible/roles/paas-proxy/files/404.json)).

To go back to the main discussion, this means point (2) of the PR description isn't right, as `__str__` should also output something like `400 - <json of errors>`. Whether that's unhelpful (point 1) is another, more subjective matter. It's still true that the Ruby client does something different, but I'm not sure it's better. Putting `__str__` to one side, what is quite unintuitive is `exc.message` returns an array, not a string. I don't feel strongly about it, though.

@lfdebrux does this change what you'd like to do here?

I think there's something to be gained by making the change we've discussed, but it's also quite a small gain, and may surprise someone who's relying on the current behaviour of `__str__` / `message`.

PR #173, opened 2020-07-16
Update documentation, including 'Get message status' section

karlchillmaid:
This PR updates the 'Get message status' section of the Python docs, along with a few other content fixes.

Specifically:

* make the use of 'message' and 'notification' more consistent
* be more specific about whether we mean 'email', 'text message' or 'letter'
* reorder the 'Get message status' section so the status descriptions come after the method
* add links to the status descriptions from the response sections, to help users understand their API response
* remove `failed` from the status tables and move it to the arguments
* update the instructions for how to fix the 404 get status error
* update descriptions of the retention period to reflect that some teams have >7 days
* improve the formatting of 400 error statuses
* update and improve message status descriptions – especially for letters and precompiled letters
* replace the word 'omit' with something simpler


DOCUMENTATION.md
```diff
@@ -734,6 +698,42 @@ If the request is not successful, the client returns an `HTTPError` containing t
...
+|:---|:---|
+|Failed|The only failure status that applies to letters is `technical-failure`. GOV.UK Notify had an unexpected error while sending to our printing provider.|
+|Accepted|GOV.UK Notify has sent the letter to the provider to be printed.|
+|Received|The provider has printed and dispatched the letter.|
+
+## Precompiled letter status descriptions
+
+|Status|information|
+|:---|:---|
+|Pending virus check|GOV.UK Notify has not completed a virus scan of the precompiled letter file.|
```
karlchillmaid, line 733: Need to add any other letter statuses that apply to precompiled letters

DOCUMENTATION.md
```diff
@@ -734,6 +698,42 @@ If the request is not successful, the client returns an `HTTPError` containing t
...
+
+## Letter status descriptions
+
+|Status|information|
+|:---|:---|
+|Failed|The only failure status that applies to letters is `technical-failure`. GOV.UK Notify had an unexpected error while sending to our printing provider.|
```
karlchillmaid, line 725: Need to add any missing statuses

DOCUMENTATION.md
```diff
@@ -734,6 +698,44 @@ If the request is not successful, the client returns an `HTTPError` containing t
...
+|Delivered|The message was successfully delivered.|
+|Failed|This covers all failure statuses:<br>- `permanent-failure` - "The provider could not deliver the message. This can happen if the phone number was wrong or if the network operator rejects the message. If you’re sure that these phone numbers are correct, you should [contact GOV.UK Notify support](https://www.notifications.service.gov.uk/support). If not, you should remove them from your database. You’ll still be charged for text messages that cannot be delivered."<br>- `temporary-failure` - "The provider could not deliver the message. This can happen when the recipient’s phone is off, has no signal, or their text message inbox is full. You can try to send the message again. You’ll still be charged for text messages to phones that are not accepting messages."<br>- `technical-failure` - "Your message was not sent because there was a problem between Notify and the provider.<br>You’ll have to try sending your messages again. You will not be charged for text messages that are affected by a technical failure."|
```
karlchillmaid, line 719: @idavidmcdonald do we return `failed` as a status? Or is it always the more specific `permanent-failure`, `temporary-failure` and `technical-failure`?

idavidmcdonald:
So I've done some research on letter statuses

So the status ‘accepted’ is displayed for any letter that has been saved on our side but not yet sent to DVLA and also saved on our side and has been sent to DVLA for printing (but not yet marked as ‘delivered’). There is no ‘created’ status before we send to DVLA, we show it as ‘accepted’ in either case. Therefore, I think the description of 'accepted' in the table could updated to be more accurate.

I also would suggest making the table in a more useful order so accepted, received, failed. People are very unlikely to get a technical failure on a letter.

Precompiled letters goes pending virus check, (potentially) failed virus check, (potentially) failed validation, accepted, received. As mentioned, I suggest we add accepted, received and technical-failure to that table.

Also, it  appears to be no 'returned' letter status exposed. We may mark a letter as 'returned' in the DB but then we still continue to tell the user that it is 'received'

DOCUMENTATION.md
idavidmcdonald: It's always more specific

DOCUMENTATION.md
idavidmcdonald: As per my larger comment earlier describing all the failure states, returned is not currently a letter state we expose so likely suggest removing that row

DOCUMENTATION.md
idavidmcdonald: See my other comment but to add created, delivered, failed (technical-failure)

DOCUMENTATION.md
karlchillmaid: In that case I might split these out in the table, instead of grouping them all under 'Failed' in the status column. This moves us closer to being able to style the statuses as code without being inaccurate or misleading.

DOCUMENTATION.md
```diff
@@ -734,11 +698,56 @@ If the request is not successful, the client returns an `HTTPError` containing t
...
+|`technical-failure`|Your message was not sent because there was a problem between Notify and the provider.<br>You’ll have to try sending your messages again.|
+
+## Text message status descriptions
+
+|Status|Description|
+|:---|:---|
+|`created`|GOV.UK Notify has placed the message in a queue, ready to be sent to the provider. It should only remain in this state for a few seconds.|
+|`sending`|GOV.UK Notify has sent the message to the provider. The provider will try to deliver the message to the recipient for up to 72 hours. GOV.UK Notify is waiting for delivery information.|
+|`pending`|GOV.UK Notify is waiting for more delivery information.<br>GOV.UK Notify received a callback from the provider but the recipient’s device has not yet responded. Another callback from the provider determines the final status of the text message.|
+|`sent` or <br>`sent-internationally`|The message was sent to an international number. The mobile networks in some countries do not provide any more delivery information. The GOV.UK Notify client API returns this status as `sent`. The GOV.UK Notify client app returns this status as `Sent internationally`.|
```
karlchillmaid, line 719: is `sent-internationally` an API status?

DOCUMENTATION.md
```diff
@@ -734,11 +698,56 @@ If the request is not successful, the client returns an `HTTPError` containing t
...
+
+## Precompiled letter status descriptions
+
+|Status|Description|
+|:---|:---|
+|`pending-virus-check`|GOV.UK Notify has not completed a virus scan of the precompiled letter file.|
+|`virus-scan-failed`|GOV.UK Notify found a potential virus in the precompiled letter file.|
+|`validation-failed`|Content in the precompiled letter file is outside the printable area. See the [GOV.UK Notify PDF letter specification](https://docs.notifications.service.gov.uk/documentation/images/notify-pdf-letter-spec-v2.3.pdf) for more information.|
+|`accepted`|GOV.UK Notify has sent the letter to the provider to be printed.|
+|`cancelled`|Sending cancelled. The letter will not be printed or dispatched.|
```
karlchillmaid, line 742: Is `cancelled` a valid state for a precompiled letter?

DOCUMENTATION.md
```diff
@@ -734,11 +698,56 @@ If the request is not successful, the client returns an `HTTPError` containing t
...
+|`technical-failure`|Your message was not sent because there was a problem between Notify and the provider.<br>You’ll have to try sending your messages again.|
+
+## Text message status descriptions
+
+|Status|Description|
+|:---|:---|
+|`created`|GOV.UK Notify has placed the message in a queue, ready to be sent to the provider. It should only remain in this state for a few seconds.|
+|`sending`|GOV.UK Notify has sent the message to the provider. The provider will try to deliver the message to the recipient for up to 72 hours. GOV.UK Notify is waiting for delivery information.|
+|`pending`|GOV.UK Notify is waiting for more delivery information.<br>GOV.UK Notify received a callback from the provider but the recipient’s device has not yet responded. Another callback from the provider determines the final status of the text message.|
+|`sent` or <br>Sent internationally|The message was sent to an international number. The mobile networks in some countries do not provide any more delivery information. The GOV.UK Notify client API returns this status as `sent`. The GOV.UK Notify client app returns this status as 'Sent internationally'.|
```
karlchillmaid, line 719: This might be clearer if we removed 'Sent internationally` from the status column – assuming that's not a status the API would ever return.

DOCUMENTATION.md
idavidmcdonald: This will need looking into.

It appears that it is correct that - The GOV.UK Notify client API returns this status as `sent`

However, I can't see anything in the python api client at least that transforms that into 'Sent internationally' or 'sent-internationally' but I would like a second opinion on this please as I'm finishing up before I go away and don't have time to double check

DOCUMENTATION.md
idavidmcdonald: Yes, that appears to be the case

DOCUMENTATION.md
```diff
@@ -728,17 +692,59 @@ If the request is not successful, the client returns an `HTTPError` containing t
 
 |error.status_code|error.message|How to fix|
 |:---|:---|:---|
-|`400`|`[{`<br>`"error": "ValidationError",`<br>`"message": "bad status is not one of [created, sending, delivered, pending, failed, technical-failure, temporary-failure, permanent-failure]"`<br>`}]`|Contact the GOV.UK Notify team|
-|`400`|`[{`<br>`"error": "ValidationError",`<br>`"message": "Apple is not one of [sms, email, letter]"`<br>`}]`|Contact the GOV.UK Notify team|
-|`403`|`[{`<br>`"error": "AuthError",`<br>`"message": "Error: Your system clock must be accurate to within 30 seconds"`<br>`}]`|Check your system clock|
-|`403`|`[{`<br>`"error": "AuthError",`<br>`"message": "Invalid token: API key not found"`<br>`}]`|Use the correct API key. Refer to [API keys](#api-keys) for more information|
+|`400`|`[{`<br>`"error": "ValidationError",`<br>`"message": "bad status is not one of [created, sending, delivered, pending, failed, technical-failure, temporary-failure, permanent-failure]"`<br>`}]`|Change or leave out the [status argument](#status-optional)|
```
karlchillmaid, line 695: Is this a helpful format for the _How to fix_ message?

DOCUMENTATION.md
leohemsted: we never return `failed` as a status. However, when fetching notifications, you can set `status='failed'` to return all failed notifications, regardless of whether they were permanent, technical or temporary. 

If you were building up a failed messages dashboard or report, you should use this otherwise you'd get, eg, "the 250 most recent temporary failures" and "the 250 most recent permanent failures" separately and the time spans of those two might not match, leading to confusing results.

At least three different services have used this search filter a total of ~1500 times in the last week

karlchillmaid: @leohemsted That's really good to know.

As these tables are intended to explain each status – is there value in doing something like this?

|Status|Description|
|:---|:---|
|`failed`| This covers all 3 failure statuses: `permanent-failure`, `temporary-failure` and `technical-failure`|
|`permanent-failure`|The provider could not...|
|`temporary-failure`|The provider could not...|
|`technical-failure`|Your message was not...|

DOCUMENTATION.md
karlchillmaid: ➕  to changing "The GOV.UK Notify client app".

I'm speaking to CHS about this – as we should come up with a name we can use consistently in the docs, on GOV.UK Notify and amongst ourselves as a team.

DOCUMENTATION.md
karlchillmaid: Is it unhelpful to include 'leave out' for both status *and* type?

leohemsted: yeah probably

DOCUMENTATION.md
servingUpAces: Let's remove it.

DOCUMENTATION.md
karlchillmaid: We've decided on 'GOV.UK Notify website' to describe the UI. This term should be used universally in the docs, on Notify and in our correspondence. I'll update the style guide in our team wiki to reflect this.

DOCUMENTATION.md
```diff
@@ -722,23 +693,74 @@ If the request to the client is successful, the client returns a `dict`.
...
 
 If the request is not successful, the client returns an `HTTPError` containing the relevant error code:
 
 |error.status_code|error.message|How to fix|
 |:---|:---|:---|
-|`400`|`[{`<br>`"error": "ValidationError",`<br>`"message": "bad status is not one of [created, sending, delivered, pending, failed, technical-failure, temporary-failure, permanent-failure]"`<br>`}]`|Contact the GOV.UK Notify team|
-|`400`|`[{`<br>`"error": "ValidationError",`<br>`"message": "Apple is not one of [sms, email, letter]"`<br>`}]`|Contact the GOV.UK Notify team|
-|`403`|`[{`<br>`"error": "AuthError",`<br>`"message": "Error: Your system clock must be accurate to within 30 seconds"`<br>`}]`|Check your system clock|
-|`403`|`[{`<br>`"error": "AuthError",`<br>`"message": "Invalid token: API key not found"`<br>`}]`|Use the correct API key. Refer to [API keys](#api-keys) for more information|
+|`400`|`[{`<br>`"error": "ValidationError",`<br>`"message": "‘bad status’ is not one of [created, sending, delivered, pending, failed, technical-failure, temporary-failure, permanent-failure]"`<br>`}]`|Change the [status argument](#status-optional)|
```
karlchillmaid, line 709: @servingUpAces does this error message accurately reflect the list of statuses given in the API?

DOCUMENTATION.md
```diff
@@ -722,23 +693,74 @@ If the request to the client is successful, the client returns a `dict`.
...
+## Text message status descriptions
+
+|Status|Description|
+|:---|:---|
+|`created`|GOV.UK Notify has placed the message in a queue, ready to be sent to the provider. It should only remain in this state for a few seconds.|
+|`sending`|GOV.UK Notify has sent the message to the provider. The provider will try to deliver the message to the recipient for up to 72 hours. GOV.UK Notify is waiting for delivery information.|
+|`pending`|GOV.UK Notify is waiting for more delivery information.<br>GOV.UK Notify received a callback from the provider but the recipient’s device has not yet responded. Another callback from the provider determines the final status of the text message.|
+|`sent`|The message was sent to an international number. The mobile networks in some countries do not provide any more delivery information. The GOV.UK Notify website displays this status as 'Sent internationally'.|
+|`delivered`|The message was successfully delivered.|
+|`failed`| This covers all 3 failure statuses: `permanent-failure`, `temporary-failure` and `technical-failure`.|
```
idavidmcdonald, line 733: From what I can see, `failed` is a deprecated status so I don't think we will return this anymore from the API? And therefore it wouldn't make sense to have a row for `failed` in any of the status tables. Is that right @servingUpAces 

https://github.com/alphagov/notifications-api/blob/32f5f454dec31b540e39e27f9fd845298f15ae2b/app/models.py#L1466

karlchillmaid: After talking to Leo and David, we removed the `failed` status from the status description tables because:

* this is not a status that will be returned by the API
* the purpose of the tables is to describe statuses that can be returned by the API

We moved `failed` to the arguments for How to get the status of multiple messages, for users who want to know how to return all failure statuses.

PR #175, opened 2020-08-17
Bump python version

leohemsted:
### remove docker folder

it's just duplicating the dockerfile in the root directory and we don't need it

### test against python 3.7 instead of 3.5

really we'd like to test against 3.6 (as that's what we use, at least), however, need to bump to 3.7.
The reason we're bumping to 3.7 is that python3-dev and gcc are required to build the psutil library from source


Dockerfile
```diff
@@ -11,16 +11,20 @@ RUN \
...
-		python3.5 \
+		python3.7 \
 		python3-pip \
 		python3-setuptools \
 		python-pip \
 		python-setuptools \
 	&& echo "Clean up" \
 	&& rm -rf /tmp/*
 
+RUN \
```
idavidmcdonald, line 22: Should be merged into the above statement?

PR #177, opened 2020-09-04
Update address fields to include line 7

karlchillmaid:
The client documentation talks about address lines 1-6 and postcode (including mandatory nature of address lines 1,2 and postcode). Now that we are supporting address lines 1-7 we need to update the documentation to reflect this.

There is a question about what we should say in the docs about the mapping that we do from postcode, or that it is sort of degraded, but needs a bit of a thought for that.

After we have done this story, services will see the new address personalisation that we require and understand that it won’t be a breaking change for them.


DOCUMENTATION.md
```diff
@@ -406,7 +410,7 @@ If the request is not successful, the client returns an `HTTPError` containing t
 |`400`|`[{`<br>`"error": "BadRequestError",`<br>`"message": "Cannot send letters with a team api key"`<br>`}]`|Use the correct type of [API key](#api-keys)|
 |`400`|`[{`<br>`"error": "BadRequestError",`<br>`"message": "Cannot send letters when service is in trial mode - see https://www.notifications.service.gov.uk/trial-mode"`<br>`}]`|Your service cannot send this letter in  [trial mode](https://www.notifications.service.gov.uk/features/using-notify#trial-mode)|
 |`400`|`[{`<br>`"error": "ValidationError",`<br>`"message": "personalisation address_line_1 is a required property"`<br>`}]`|Ensure that your template has a field for the first line of the address, check [personalisation](#send-a-letter-arguments-personalisation-optional) for more information.|
-|`400`|`[{`<br>`"error": "ValidationError",`<br>`"message": "Must be a real UK postcode"`<br>`}]`|Ensure that the value for the postcode field in your letter is a real UK postcode|
+|`400`|`[{`<br>`"error": "ValidationError",`<br>`"message": "Must be a real UK postcode"`<br>`}]`|Ensure that the value for the address_line_7 field in your letter is a real UK postcode or the name of a country outside the UK|
```
karlchillmaid, line 413: Are we updating the message in the API?

DOCUMENTATION.md
```diff
@@ -511,7 +515,7 @@ If the request to the client is successful, the client will return a `dict`:
   "line_4": "ADDRESS LINE 4", # optional string for letter
   "line_5": "ADDRESS LINE 5", # optional string for letter
   "line_6": "ADDRESS LINE 6", # optional string for letter
-  "postcode": "A REAL UK POSTCODE", # required string for letter
+  "line_7": "A REAL UK POSTCODE", # required string for letter
```
karlchillmaid, line 518: do we need to add something about country name here?

DOCUMENTATION.md
```diff
@@ -340,15 +340,19 @@ The personalisation argument always contains the following required parameters f
 
 - `address_line_1`
 - `address_line_2`
-- `postcode` (this needs to be a real UK postcode)
+– `address_line_7` this needs to be a real UK postcode or the name of a country outside the UK
+
+Notify uses `address_line_7` to check for international addresses. This is so we can charge you the correct postage.
```
karlchillmaid, line 345: Is it useful to have this context?

DOCUMENTATION.md
```diff
@@ -340,15 +340,19 @@ The personalisation argument always contains the following required parameters f
...
+
+The `postcode` personalisation argument has been replaced. If your template still uses `postcode`, Notify will treat it as `address_line_7`.
```
karlchillmaid, line 347: "treat it as"?
"read it as"?
"map it to"?

What's the best way to describe this for our users?

DOCUMENTATION.md
leohemsted: if the service has international letters enabled as a permission then the API returns "Last line of address must be a real UK postcode or another country". 

if they haven't got that setting, then "Must be a real UK postcode".

i'm not sure if we are turning international letters permissions on by default?

DOCUMENTATION.md
leohemsted: i'd probably just revert this to "ADDRESS LINE 7"

DOCUMENTATION.md
leohemsted: yes but no - we don't just look at address line 7. We look at the very last non-empty line. For example, consider the following address:

address_line_1: "123 champs elysees"
address_line_2: "paris"
address_line_3: "12345"
address_line_4: "FRANCE"
address_line_5: ""
address_line_6: ""
address_line_7: ""

we'd check that the last non empty line has a correct country or UK postcode

karlchillmaid: Cool. So we *can* still read that as a valid international address, but we want to encourage users to use line_7?

DOCUMENTATION.md
```diff
@@ -336,19 +336,35 @@ To find the template ID:
...
 
 ```python
 personalisation={
   'address_line_1': 'The Occupier',
   'address_line_2': '123 High Street',
-  'postcode': 'SW14 6BF',
+  'address_line_3': 'Richmond upon Thames',
+  'address_line_4': 'Middlesex',
+  'address_line_5': 'SW14 6BF',
+  'address_line_6': '',
```
karlchillmaid, line 366: Should I leave out lines 6 and 7 in this example, as they're not populated?

servingUpAces (approved):

DOCUMENTATION.md
```diff
@@ -336,19 +336,35 @@ To find the template ID:
...
+– `address_line_6`
+– `address_line_7`
+
+The address must have at least 3 lines.
+
+The last line needs to be a real UK postcode or the name of a country outside the UK.
+
+Notify checks for international addresses and will automatically charge you the correct postage.
+
+The `postcode` personalisation argument has been replaced. If your template still uses `postcode`, Notify will treat it as the last line of the address.
```
leohemsted, line 355: this one still feels a little weird (cos its explaining such a weird legacy edge casey thing, i suppose), but i think this is fine for now

PR #180, opened 2020-09-16
Fix broken internal links and other minor changes

klssmith:
* Fixes broken internal links
* Corrects one use of the wrong PDF letter template
* Makes all the text that links to the Notify sign in page the same for consistency


tombye (approved):
Looks good to me 👍

PR #185, opened 2021-01-04
Remove support for Python 2.7

leohemsted:

Drop Python 2.7 Support. Python 2.7 is now end of life, and has been for a while. Instead, test against 3.6 through 3.9

We were previously just calling `apt-get install python python3 [...]`. This isn't a scalable approach when we need to install multiple python 3 environments, so switch to using pyenv. This PR goes slightly against the typical wisdom of putting `.python-version` files in `.gitignore`, however, it does provide a very convenient way of defining the exact versions of python for pyenv to fetch. We should make sure we bump the versions in this file every now and then to ensure that we're testing against the most recent patches of each major python version.

I'm trying to work out if doing this clashes with existing virtual environments.

## Checklist

  - [x] `DOCUMENTATION.md`
  - [x] `CHANGELOG.md`
  - [ ] `CONTRIBUTING.md`


CrystalPea (approved):
Good work, and works well for me 👍🏼

PR #197, opened 2021-06-11
Optionally include notifications from jobs

benthorner:
https://govuk.zendesk.com/agent/tickets/4590242

Previously we would exclude these. We have a service [1] that:

- Uses the API to send initially.
- Uses the API to query failures.
- Uses the UI to manually resend.
- Uses the API to check the results.

While I'm not entirely convinced the last step is necessary, it
_is_ confhsing that what's visible in the UI is inconsistent with
the result from the API client. So, for at least that reason, this
adds the option to include notifications sent via a job.

Note that our Ruby client implicitly allows this already, by not
restricting the set of allowed parameters [2].

[1]: 53be32cd-69c1-4a72-a2e8-46ad8f768c61
[2]: https://github.com/alphagov/notifications-ruby-client/blob/328362a2df82cc0cb142618065467a032416cbd3/lib/notifications/client.rb#L101

## Checklist

  - [x] `DOCUMENTATION.md`
  - [x] `CHANGELOG.md`

Separately:

- [ ] Let the user on the ticket know the change is being shipped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 6.2.0
+
+* Optionally include job notifications in `get_all_notifications`
+
 ## 6.1.0
 
 * Add type hints for errors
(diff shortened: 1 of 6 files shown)


servingUpAces (approved):

benthorner:
Looks like we need to update the Java client as well. I'll try and do that soon. 

For now, the user has confirmed they'll get some benefit from the change here.

PR #199, opened 2021-06-17
Fix anchor link not working for new argument

benthorner:
## Checklist

  - [ ] `DOCUMENTATION.md`
  - [ ] `CHANGELOG.md`
  - [ ] `CONTRIBUTING.md`

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -574,7 +574,7 @@ You can filter the returned messages by including the following optional argumen
 - [`status`](#status-optional)
 - [`reference`](#get-the-status-of-multiple-messages-arguments-reference-optional)
 - [`older_than`](#older-than-optional)
-- [`include_jobs`](#include-jobs)
+- [`include_jobs`](#include-jobs-optional)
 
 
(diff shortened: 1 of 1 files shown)


idavidmcdonald (approved):

idavidmcdonald:
Worth just mentioning here that I don't know if we expose the word `jobs` to our users much, if not, at all. We potentially call them `uploads` but I don't know if we've been able to describe exactly what this does. `jobs` I think is a word we just use for our code.

Ship might have sailed on that given this has been merged but just thought I'd flag to @karlchillmaid just in case.

benthorner:
I was wondering about that, which is why I wrote the doc a bit differently:

> Includes notifications sent as part of a batch upload.
>
> If you leave out this argument, the method only returns notifications sent using the API.

The name of the argument is just so it matches what we have in the API, which also makes it consistent with the Ruby client (for anyone who's passing it into the unrestricted list of argument we allow over there).

benthorner:
...add this is where I start to regret doing this pro bono: https://github.com/alphagov/notifications-python-client/pull/200

PR #201, opened 2021-07-15
Support encoding sets in JSON POST data

benthorner:
https://www.pivotaltracker.com/story/show/178770155

Sets are just a special kind of list, so there's no reason why we
can't support them in POST data. Rather than peppering our code with
list(set) conversions (and testing them), it seems more pragmatic to
do it once where we can get the most benefit from it.

One use case for this will be POSTing permissions from Admin to API.
Admin uses sets for permissions most of the time, which means we get
an error if we try to POST them directly:

      File "/Users/benthorner/.pyenv/versions/3.6.12/lib/python3.6/json/encoder.py", line 180, in default
        o.__class__.__name__)
      TypeError: Object of type 'set' is not JSON serializable

## Checklist

  - [ ] `DOCUMENTATION.md`
  - [ ] `CHANGELOG.md`
  - [ ] `CONTRIBUTING.md`

### DIFF
--- a/notifications_python_client/base.py
+++ b/notifications_python_client/base.py
@@ -74,7 +74,7 @@ class BaseAPIClient(object):
         }
 
         if data is not None:
-            kwargs.update(data=json.dumps(data))
+            kwargs.update(data=self._serialize_data(data))
 
         if params is not None:
             kwargs.update(params=params)
@@ -82,6 +82,15 @@ class BaseAPIClient(object):
         url = urllib.parse.urljoin(str(self.base_url), str(url))
 
         return url, kwargs
+
+    def _serialize_data(self, data):
+        return json.dumps(data, default=self._extended_json_encoder)
+
+    def _extended_json_encoder(self, obj):
+        if isinstance(obj, set):
+            return list(obj)
+
+        raise TypeError
 
     def _perform_request(self, method, url, kwargs):
         start_time = time.monotonic()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 6.2.1
+
+* Auto-convert sets to lists in API calls
+
 ## 6.2.0
 
 * Optionally include job notifications in `get_all_notifications`
--- a/notifications_python_client/__init__.py
+++ b/notifications_python_client/__init__.py
@@ -7,7 +7,7 @@
 #
 # -- http://semver.org/
 
-__version__ = '6.2.0'
+__version__ = '6.2.1'
 
 from notifications_python_client.errors import REQUEST_ERROR_STATUS_CODE, REQUEST_ERROR_MESSAGE  # noqa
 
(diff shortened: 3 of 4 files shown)


notifications_python_client/base.py
```diff
@@ -83,6 +83,16 @@ def _create_request_objects(self, url, data, params):
 
         return url, kwargs
 
+    def _serialize_data(self, data):
+        class ExtendedJSONEncoder(json.JSONEncoder):
```
quis, line 87: Is there a good reasons for re-defining the class each time the method is called?

benthorner: Hmm...the redefining didn't occur to me.

I wanted to keep the class close to the one place where it's used. It's also not meant to be part of the module namespace, so hiding it away in here seemed like a good solution.

Do you think there's a performance impact if we redefine it each time? One alternative might be to just move it directly above the method, so at least it's still nearby. How about that?

quis: I just did a toy example and found no performance difference.

<details><summary>Example A</summary>

```python

class Foo:

    def __init__(self):
        pass

    def foo(self):

        class Bar:

            def __init__(self):
                pass

            def bar():
                return 1 + 2

        return Bar().bar()

for i in range(1_000_000):
    Foo().foo()
```
</details>

<details><summary>Example B</summary>

```python

class Bar:

    def __init__(self):
        pass

    def bar():
        return 1 + 2

class Foo:

    def __init__(self):
        pass

    def foo(self):

        return Bar().bar()

for i in range(1_000_000):
    Foo().foo()
```
</details>

I think it’s more that when I see unpythonic<sup>1</sup> code like that I go ‘hmm, I wonder what edge case this developer was working around that meant they had to nest a class inside a method’. Whereas if you put the class in its own module I don’t have to wonder.

***

1. The [Zen of Python](https://www.python.org/dev/peps/pep-0020/) says:
    > Simple is better than complex.
    >
    > Flat is better than nested.

benthorner: Thanks for checking the performance 🏅.

(I'd argue Python isn't as Pythonic as it used to be.)

It sounds like you'd find it less surprising to see this class in a separate module. Conversely, I'd find it more surprising: to me that implies it's "the" JSON encoder we use "for any JSON". I don't feel that strongly about it, but I was thinking a compromise of moving it just outside the method might be better all round.

Would you prefer it to be its own module, or would decreasing the nesting be OK?

quis: As far as I can see this is the only place we encode JSON in this repo therefore it is "the" JSON encoder for the Python client. I can’t see a future where we’d want the client using two different encoders.

Decreasing the nesting is OK though, since that makes it clear there’s no weird edge case that means the class needs to be redefined each time the method is called.

benthorner: OK I've amended the commit after also reading [the manual](https://docs.python.org/3/library/json.html#basic-usage) (😄) and realising there's a simpler way of doing it anyway, without having to define a class.

quis: > If specified, default should be a function that gets called for objects that can’t otherwise be serialized. It should return a JSON encodable version of the object or raise a TypeError. If not specified, TypeError is raised.

👍🏻

quis (approved):

PR #190, opened 2021-03-18
Update allowed formats for file uploads

benthorner:
This follows the change in [1].

[1]: https://github.com/alphagov/notifications-tech-docs/commit/f4cfbed16a1e2a37f8a5553adb2a65b82414443f

## Checklist

- [x] I’ve used the pull request template
- [ ] I’ve written unit tests for these changes
- [ ] I’ve update the documentation in
  - [ ] `DOCUMENTATION.md`
  - [ ] `CHANGELOG.md`
- [ ] I’ve bumped the version number in
  - [ ] `notifications_python_client/__init__.py`
  - [ ] `notifications_python_client/test_base_api_client.py`, function `test_user_agent_is_set`
- [ ] I've added new environment variables to
  - [ ] `notifications-python-client/scripts/generate_docker_env.sh`
  - [ ] `notifications-python-client/tox.ini`
  - [ ] `CONTRIBUTING.md`

PR #203, opened 2021-08-13
Update documentation - add xlsx to supported document upload types

CrystalPea:


klssmith (approved):

klssmith:
Having just looked at all these docs PRs, I've just noticed that @benthorner took a different approach to the formatting https://github.com/alphagov/notifications-python-client/pull/190 so you might want to discuss that with him or consider that way instead.

PR #204, opened 2021-09-15
Add default timeout to base client

jzbahrai:
## Checklist

  - [ ] `DOCUMENTATION.md`
  - [ ] `CHANGELOG.md`
  - [ ] `CONTRIBUTING.md`


tests/notifications_python_client/test_base_api_client.py
```diff
@@ -9,8 +9,8 @@
 
 
 @pytest.mark.parametrize('client', [
-    BaseAPIClient(api_key=COMBINED_API_KEY),
-    BaseAPIClient(COMBINED_API_KEY),
+    BaseAPIClient(api_key=COMBINED_API_KEY, timeout=10),
+    BaseAPIClient(COMBINED_API_KEY, timeout=10),
```
benthorner, line 13: This is unrelated to the purpose of this test: checking we pass through `service_id` and `key`. We're also not checking `timeout` is adopted correctly, which you could do with `mocker` or just test the value is at least set on the client object instance.

How about checking this in its own test?

benthorner (commented):
Thanks for raising this @jzbahrai ⭐.

I've got one comment where you could improve the test so that it covers the new feature more clearly. 

We also talked about this as a team and our conclusion was the new default is fine. There was some discussion about whether to [specify a separate `read_timeout` and `connect_timeout`](https://docs.python-requests.org/en/latest/user/advanced/#timeouts). I don't feel too strongly about this myself as 30s for both is still an improvement over "infinity" and the optional parameter allows someone to specify a tuple of separate values if they like.

tests/notifications_python_client/test_base_api_client.py
jzbahrai: Thank you so much for responding! I updated the tests to check that the default time out is 30/ when it is set it is the correct value on the instance.

benthorner (commented):
🚀  Thanks for fixing up the tests @jzbahrai !

⚠️  I've run this through our CI machine and my clone PR is green. Unfortunately something is preventing me from getting our CI to run on this original PR (our normal process for doesn't seem to be working). Would you be OK with me cloning this PR and merging that? I'll be sure to refer to this one so it's clear who did the work.

💬  When I was re-reading the diff I noticed one more issue that I think is worth addressing. I was tempted to call it optional, and if you'd rather I can make a separate commit to address it in my clone PR. Let me know what you'd like to do.

benthorner:
@jzbahrai no problem 👍.

Closing in favour of https://github.com/alphagov/notifications-python-client/pull/211.

Issue #205, opened 2021-10-01
PyPI package includes integration_test as top-level package


reupen:
Hi,

I noticed that the PyPI package includes `integration_test` as a top-level package.

This means that after installing the package (e.g. `pip install notifications-python-client`) you can do things like:

```python
import integration_test
from integration_test.integration_tests import get_template_by_id
```

This isn't causing me a problem, but it could lead to accidental imports or conflicts etc. (as `integration_test` is a pretty generic package name).

(In my case, I ended up there when searching for uses of `NotificationsAPIClient` in my IDE, and was a bit confused for a second...)

quis:
Hey, thanks for taking the time to raise the issues. We recently fixed this in one of our other packages so I’ve raised [a pull request](https://github.com/alphagov/notifications-python-client/pull/206/files) with the same fix for this package.

ushkarev:
I've just noticed that this package still packages a bunch of test files that are not relevant and that #206 has been closed. Do you perhaps want a PR with a fix?

reupen:
> I've just noticed that this package still packages a bunch of test files that are not relevant 

Indeed, it installs both `tests` and `integration_test` packages.

ushkarev:
…and looking in the latest wheel, `utils` too!

samuelhwilliams:
🙇‍♂️  Thanks for putting up this PR @ushkarev - we will try to take a look at this soon with a view to getting it merged 🎉

PR #206, opened 2021-10-01
Only include source in setup.py

quis:
Copies the approach taken in https://github.com/alphagov/notifications-utils/pull/898

It's unlikely we'll add any separate packages, and this prevents us accidentally bundling the integration tests folder in.

This was previously resulting in a tests folder being found in the site_packages in a virtualenv, containing all the integration tests from the Python client in every venv that required the Python client repo.

Fixes https://github.com/alphagov/notifications-python-client/issues/205

## Checklist

  - [ ] `DOCUMENTATION.md`
  - [ ] `CHANGELOG.md`
  - [ ] `CONTRIBUTING.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 6.3.1
+
+* Remove test directories from distributed package
+
 ## 6.3.0
 
 * Added a default timeout of 30s to the `BaseAPIClient`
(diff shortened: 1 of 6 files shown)


idavidmcdonald:
Will need a version bump? I think we could get away with a patch give no one should be using it? Or do we play it super safe and go for a major even though it seems overcautious?

samuelhwilliams:
Do we still want to get this in?

PR #215, opened 2022-08-26
Add details of ‘confirm email address’ and ‘custom retention period’

karlchillmaid:
# Do not merge until the new features are live

This PR updates the ‘Send a file by email’ section of the REST-API documentation.

There are 2 stories in the backlog which describe this work:

* [Update 'send a file by email' API client docs for 'email verification' flow](https://www.pivotaltracker.com/story/show/182942256)
* [Update 'send a file by email' API client docs for custom retention periods](https://www.pivotaltracker.com/story/show/182999152)

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -286,7 +286,7 @@ When a recipient clicks the link in the email you’ve sent them, they have to e
 
 From 29 March 2023, we will turn this feature on by default for every file you send. Files sent before 29 March will not be affected.
 
-##### Opt in
+##### Turn on email address check
 
 To ask recipients to confirm their email address, set the `verify_email_before_download` flag to `true`.
 
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -308,7 +308,6 @@ with open('file.pdf', 'rb') as f:
 
 You should not opt out if you send files that contain personally identifiable information or sensitive information.
 
-If you do not want to use this security feature, set the `verify_email_before_download` flag to `false` by 29 March 2023.
 
 ```python
 from notifications_python_client import prepare_upload
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -289,8 +289,8 @@ From 29 March 2023, we will turn this feature on by default for every file you s
 ##### Turn on email address check
 
 To use this feature before 29 March 2023 you will need version X.X.X of the Python client library, or a more recent version.
- 
-To make the recipient confirm their email address before downloading the file, set the `verify_email_before_download` flag to `True`.
+
+To make the recipient confirm their email address before downloading the file, set the `confirm_email_before_download` flag to `True`.
 
 You will not need to do this after 29 March.
 
@@ -302,7 +302,7 @@ with open('file.pdf', 'rb') as f:
     personalisation={
       'first_name': 'Amala',
       'application_date': '2018-01-01',
-      'link_to_file': prepare_upload(f, verify_email_before_download=True),
+      'link_to_file': prepare_upload(f, confirm_email_before_download=True),
     }
 ```
 
@@ -318,7 +318,7 @@ You should not turn this feature off if you send files that contain:
 * commercially sensitive information
 * information classified as ‘OFFICIAL’ or ‘OFFICIAL-SENSITIVE’ under the [Government Security Classifications](https://www.gov.uk/government/publications/government-security-classifications) policy
 
-To let the recipient download the file without confirming their email address, set the `verify_email_before_download` flag to `False`.
+To let the recipient download the file without confirming their email address, set the `confirm_email_before_download` flag to `False`.
 
 
 ```python
@@ -329,7 +329,7 @@ with open('file.pdf', 'rb') as f:
     personalisation={
       'first_name': 'Amala',
       'application_date': '2018-01-01',
-      'link_to_file': prepare_upload(f, verify_email_before_download=False),
+      'link_to_file': prepare_upload(f, confirm_email_before_download=False),
     }
 ```
 
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -218,7 +218,7 @@ To send a file by email, add a placeholder to the template then upload a file. T
 
 The links are unique and unguessable. GOV.UK Notify cannot access or decrypt your file.
 
-Your file will be available to download for a default period of 78 weeks (18 months). From 29 March 2023 we will reduce this to 26 weeks (6 months) for all new files. Files sent before 29 March are not affected.
+Your file will be available to download for a default period of 78 weeks (18 months). From 29 March 2023 we will reduce this to 26 weeks (6 months) for all new files. Files sent before 29 March will not be affected.
 
 To help protect your files you can also:
 
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -280,7 +280,11 @@ with open('file.csv', 'rb') as f:
 
 #### Ask recipients to confirm their email address before they can download the file
 
-This feature is currently opt-in only. From 29 March 2023 it will apply to all new files by default, unless you opt out.
+This new security feature is optional. You should use it if you send files that are sensitive - for example, because they contain personal information about your users.
+
+When a recipient clicks the link in the email you’ve sent them, they have to enter their email address. Only someone who knows the recipient’s email address can download the file.
+
+From 29 March 2023, we will turn this feature on by default for every file you send. Files sent before 29 March will not be affected.
 
 ##### Opt in
 
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -304,7 +304,7 @@ with open('file.pdf', 'rb') as f:
     }
 ```
 
-##### Opt out (not recommended)
+##### Turn off email address check (not recommended)
 
 You should not opt out if you send files that contain personally identifiable information or sensitive information.
 
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -225,7 +225,7 @@ To help protect your files you can also:
 * ask recipients to confirm their email address before downloading
 * choose the length of time that a file is available to download
 
-To turn these features on or off, you will need to update your API client to the latest version.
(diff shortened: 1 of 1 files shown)


DOCUMENTATION.md
```diff
@@ -271,6 +275,66 @@ with open('file.csv', 'rb') as f:
...
+
+```python
+from notifications_python_client import prepare_upload
+
+with open('file.pdf', 'rb') as f:
+    ...
+    personalisation={
+      'first_name': 'Amala',
+      'application_date': '2018-01-01',
+      'link_to_file': prepare_upload(f),
```
karlchillmaid, line 296:
```suggestion
      'link_to_file': prepare_upload(f),
      'verify_email_before_download': true,
```

DOCUMENTATION.md
```diff
@@ -271,6 +275,66 @@ with open('file.csv', 'rb') as f:
...
```
karlchillmaid, line 314:
```suggestion
      'link_to_file': prepare_upload(f),
      'verify_email_before_download': false,
```

DOCUMENTATION.md
```diff
@@ -271,6 +275,66 @@ with open('file.csv', 'rb') as f:
...
```
karlchillmaid, line 334:
```suggestion
      'link_to_file': prepare_upload(f),
      'retention_period': '4 weeks',
```

DOCUMENTATION.md
samuelhwilliams: Same here:

```suggestion
      'link_to_file': prepare_upload(f),
      'verify_email_before_download': False,
```

DOCUMENTATION.md
samuelhwilliams: And here:

```suggestion
      'link_to_file': prepare_upload(f, retention_period='4 weeks'),
```

DOCUMENTATION.md
```diff
@@ -271,6 +275,67 @@ with open('file.csv', 'rb') as f:
...
+      'verify_email_before_download': True,
```
karlchillmaid, line 297:
```suggestion
      'link_to_file': prepare_upload(f, verify_email_before_download=True),
```

DOCUMENTATION.md
```diff
@@ -271,6 +275,66 @@ with open('file.csv', 'rb') as f:
     }
 ```
 
+#### Ask recipients to confirm their email address before they can download the file
+
+This feature is currently opt-in only. From 29 March 2023 it will apply to all new files by default, unless you opt out.
```
karlchillmaid, line 280:
```suggestion
This new security feature is optional. You should use it if you send files that are sensitive - for example, because they contain personal information about your users.

When a recipient clicks the link in the email you’ve sent them, they have to enter their email address. Only someone who knows the recipient’s email address can download the file.

From 29 March 2023, we will turn this feature on by default for every file you send. Files sent before 29 March are not affected.
```

DOCUMENTATION.md
```diff
@@ -271,6 +275,66 @@ with open('file.csv', 'rb') as f:
...
+
+##### Opt in
```
karlchillmaid, line 282:
```suggestion
##### Turn on email address check
```

DOCUMENTATION.md
```diff
@@ -271,6 +275,66 @@ with open('file.csv', 'rb') as f:
...
+
+To ask recipients to confirm their email address, set the `verify_email_before_download` flag to `true`.
+
+You will not need to do this after 29 March 2023.
```
karlchillmaid, line 286:
```suggestion
You can turn this feature on on a per-file basis. You will not need to do this after 29 March 2023.

To make the recipient confirm their email address before downloading the file, set the `verify_email_before_download` flag to `true`.
```

DOCUMENTATION.md
```diff
@@ -271,6 +275,66 @@ with open('file.csv', 'rb') as f:
...
+with open('file.pdf', 'rb') as f:
+    ...
+    personalisation={
+      'first_name': 'Amala',
+      'application_date': '2018-01-01',
+      'link_to_file': prepare_upload(f, verify_email_before_download=True),
+    }
+```
+
+##### Opt out (not recommended)
```
karlchillmaid, line 300:
```suggestion
##### Turn off email check (not recommended)
```

DOCUMENTATION.md
```diff
@@ -271,6 +275,66 @@ with open('file.csv', 'rb') as f:
...
+
+You should not opt out if you send files that contain personally identifiable information or sensitive information.
```
karlchillmaid, line 302:
```suggestion
You can turn this feature off on a per-file basis. You should not do this if you send files that contain either:

* personally identifiable information
* commercially sensitive information
```

DOCUMENTATION.md
```diff
@@ -271,6 +275,66 @@ with open('file.csv', 'rb') as f:
...
+
+If you do not want to use this security feature, set the `verify_email_before_download` flag to `false` by 29 March 2023.
```
karlchillmaid, line 304:
```suggestion
```

DOCUMENTATION.md
samuelhwilliams: "won't be affected" vs "are not affected"?

Otherwise looks good.

DOCUMENTATION.md
samuelhwilliams: Do we drop off the '2023' second time around just because it's too repetitive?

DOCUMENTATION.md
```diff
@@ -216,10 +216,17 @@ You can leave out this argument if your service only has one reply-to email addr
...
 The links are unique and unguessable. GOV.UK Notify cannot access or decrypt your file.
 
+Your file will be available to download for a default period of 78 weeks (18 months). From 29 March 2023 we will reduce this to 26 weeks (6 months) for all new files. Files sent before 29 March are not affected.
+
+To help protect your files you can also:
+
+* ask recipients to confirm their email address before downloading
+* choose the length of time that a file is available to download
+
+To turn these features on or off, you will need to update your API client to the latest version.
```
idavidmcdonald, line 228: I would suggest we remove this sentence for two reasons

1. We don't have this precedent anywhere else of telling people they need to update their API client to take advantage of a new feature
2. At some point it means we need to remove this line as it won't be true (as users by that point may already be using a version of the client that contains this functionality)

DOCUMENTATION.md
```diff
@@ -216,10 +216,17 @@ You can leave out this argument if your service only has one reply-to email addr
 
 To send a file by email, add a placeholder to the template then upload a file. The placeholder will contain a secure link to download the file.
 
-The file will be available for the recipient to download for 18 months.
-
 The links are unique and unguessable. GOV.UK Notify cannot access or decrypt your file.
 
+Your file will be available to download for a default period of 78 weeks (18 months). From 29 March 2023 we will reduce this to 26 weeks (6 months) for all new files. Files sent before 29 March are not affected.
```
idavidmcdonald, line 221: Very small one that is optional as maybe our documentation doesn't need to be 100% precise necessarily.

Is it beneficial to say the default period is 78 weeks. Firstly, I'm guessing it may not always be exactly 78 weeks depending on how many weeks in a month? Secondly, at this point is it useful to make them start thinking in terms of weeks?

Same with the 26 weeks (6 months).

@samuelhwilliams Here is a slightly more interesting one though, what will our default retention period actually be when we change it. Will it be 26 weeks or will it be 6 months? Those could be very slightly different by a few days.

DOCUMENTATION.md
```diff
@@ -216,10 +216,17 @@ You can leave out this argument if your service only has one reply-to email addr
...
+
+To help protect your files you can also:
+
+* ask recipients to confirm their email address before downloading
```
idavidmcdonald, line 225: Optional: will this sentence need changing in 6 months time? Maybe in 6 months time I'd phrase it more like we ask recipients to confirm their email address before downloading. You can choose to turn this off. 

I think for speed of getting this merged, I think we don't make a change on this for the moment at least.

DOCUMENTATION.md
samuelhwilliams: It'll be 26 weeks. Everything will be in weeks - which I think is why we're starting to get them to think that way by how we speak in the docs.

samuelhwilliams: At the moment, documents in the S3 bucket expire after 546 days, which is 78 weeks. So that is correct.

DOCUMENTATION.md
samuelhwilliams: Alternatively we say that they need to update to at least a specific client version x.y.z.

DOCUMENTATION.md
idavidmcdonald: Same comment about updating the API client as above

idavidmcdonald: I'm interested in why we decided to include `information classified as ‘OFFICIAL’ or ‘OFFICIAL-SENSITIVE’ under the Government Security Classifications policy` please? Was this a suggestion from IA or privacy?

And should we link to the Government Security Classifications policy here?

DOCUMENTATION.md
karlchillmaid: Thanks both, I’m resolving this.

DOCUMENTATION.md
karlchillmaid: Yes, the whole section will need a little makeover in 6 months.

All the ‘turn it on’ stuff will go away, the retention period will be fixed at 26 weeks, etc.

DOCUMENTATION.md
```diff
@@ -271,6 +278,81 @@ with open('file.csv', 'rb') as f:
...
+```
+
+#### Choose the length of time that a file is available to download
+
+Set the number of weeks you want the file to be available using the `retention_period` key.
+
+You can choose any value between 1 week and 78 weeks.
+
+If you do not do this, the file will be available for the default period of 78 weeks (18 months).
+
```
karlchillmaid, line 343:
```suggestion
You can choose any value between 1 week and 78 weeks.

To use this feature will need version X.X.X of the Python client library, or a more recent version.

If you do not choose a value, the file will be available for the default period of 78 weeks (18 months).

```

samuelhwilliams:
Combined into https://github.com/alphagov/notifications-python-client/pull/216

PR #216, opened 2022-09-26
Add support for new doc download features

samuelhwilliams:
Pass through `confirm_email_before_download` and `retention_period` parameters via `prepare_upload` calls so that clients can start to explicitly opt-in or opt-out of the new security features.

Bump client to v6.4.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 6.4.0
+
+* Added support for `confirm_email_before_download` and `retention_period` security features for sending files by email.
+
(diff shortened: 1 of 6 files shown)


CrystalPea (approved):
LGTM 👍🏼

samuelhwilliams:
Updated to always send the new fields (albeit as None) - after discussion with Leo.

samuelhwilliams:
Have pulled in all of the content changes from https://github.com/alphagov/notifications-python-client/pull/215

PR #220, opened 2023-02-10
Drop support for end-of-life python 3.6 and add support & testing for 3.10 and 3.11

ushkarev:
# What problem does the pull request solve?
Officially support latest versions of python by adding testing on 3.10 and 3.11. 3.6 reached end-of-life over a year ago so support can be dropped.

# Checklist

- [x] I’ve used the pull request template
- [x] I’ve written unit tests for these changes – no new tests added but testing is now performed on new set of python versions
- [x] I’ve update the documentation in
  - [x] `DOCUMENTATION.md`
  - [x] `CHANGELOG.md`
- [ ] I’ve bumped the version number in – no, because it’s better done by the GOV.UK Notify team, though I would suggest a major version change and could be done in concert with #219
  - [ ] `notifications_python_client/__init__.py`
- [ ] I've added new environment variables to – no relevant changes
  - [ ] `notifications-python-client/scripts/generate_docker_env.sh`– I presume this is supposed to say `run_with_docker.sh`, but maybe it’s not even needed anymore?
  - [ ] `notifications-python-client/tox.ini` – no relevant changes
  - [ ] `CONTRIBUTING.md` – should this be under environment variables?

PR #223, opened 2023-02-27
Drop support for end-of-life python 3.6

samuelhwilliams:
Local copy of #220


tox-python-versions
```diff
@@ -1,4 +1,5 @@
-3.6.12
 3.7.9
 3.8.6
 3.9.0
+3.10.9
+3.11.1
```
ushkarev, line 5: I dunno how these are supposed to be chosen~ In my projects, I've never pinned the patch-level version of python.

setup.py
```diff
@@ -45,18 +45,19 @@ def run_tests(self):
...
     ],
     keywords='gds govuk notify',
 
     packages=find_packages(include=['notifications_python_client']),
     include_package_data=True,
 
-    # only support actively patched versions of python (https://devguide.python.org/devcycle/#end-of-life-branches)
-    python_requires='>=3.6',
+    # only support actively patched versions of python (https://devguide.python.org/versions/)
+    python_requires='>=3.7',
```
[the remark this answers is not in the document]
samuelhwilliams: I am OK with us removing support altogether for EOL python versions.

tox-python-versions
samuelhwilliams: Yeah I'd be inclined to only list minor versions too, will check/update.

idavidmcdonald (approved):

PR #232, opened 2023-10-13
re-use a single request session

leohemsted:
This should unlock performance benefits by utilising HTTP keep-alive to reuse one TCP connection between many requests. Quoth the docs:

[The Session object] across all requests [...] will use urllib3’s connection pooling. So if you’re making several requests to the same host, the underlying TCP connection will be reused, which can result in a significant performance increase (see HTTP persistent connection).[^1]

Tested very unscientifically locally, 500 test notifications took 39 seconds, down from 52 seconds. (25% faster)

[^1]: https://requests.readthedocs.io/en/latest/user/advanced/

## Checklist

  - [ ] `DOCUMENTATION.md`
  - [x] `CHANGELOG.md`
  - [ ] `CONTRIBUTING.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 8.1.0
+
+* Reuse the same `requests.session` between requests to increase performance through HTTP Keepalive
+
 ## 8.0.1
 
 * Some minor non-functional code reformatting.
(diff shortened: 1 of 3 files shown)


risicle:
Only suggestion I'd make is this might deserve more than a patchlevel-version bump.

leohemsted:
good point @risicle - i'd briefly considered it. on reflection, i think this is marginally more "added functionality" than "bug fix" so i've changed the bump to a minor version, but it's much of a muchness i think.

PR #242, opened 2024-07-12
Copy common configuration files from `notifications-utils`

quis:
A bunch of our test dependencies are now specified in [notifications-utils/requirements_for_test_common.txt](https://github.com/alphagov/notifications-utils/blob/f59616870b5457d5441863366d8c99af8507ca28/requirements_for_test_common.txt) so can be removed from requirements_for_test.txt

`.pyproject.toml` and `pre-commit-config.yaml` can also be copied from utils rather than being duplicated here.

This all makes it easier to keep dependencies up to date and make changes to linting rules.

Because this introduces more linting rules there are some non-functional changes to the code here, detailed in individual commits.

***

Fixes #239
Closes #240


.gitignore
```diff
@@ -22,6 +22,9 @@ var/
 *.egg-info/
 .installed.cfg
 *.egg
+requirements_for_test_common.txt
+pyproject.toml
```
leohemsted, line 26: hmmm i'm not sure about this - pyproject.toml is a replacement for setup.py, and it was on my todo list to move all the project metadata from setup.py to pyproject.toml. Not having the file in the repo source would probably get in the way of it being parsed. Note that setup.py is not deprecated - https://packaging.python.org/en/latest/discussions/setup-py-deprecated/#setup-py-deprecated - but the movement in the python ecosystem is away from it.

i think ultimately the best solution may be to do a bit more work and write a script (using tomllib https://tomlkit.readthedocs.io/en/latest/) to merge in the `[tool.*]` sections from the utils template into the rest of the data 🤔

leohemsted (commented):
i'm in favour of the actual code changes and dropping py 3.7, but i'm a bit wary about bringing in utils and the copying-across-files-from-other-repo pattern into a library we distribute widely (and also regularly see external development on)

.gitignore
quis: What if we keep `pyproject.toml` in version control for now, but still copy it across from utils? I’d support doing the work to parse and merge later, if we decide to move away from `setup.py`

leohemsted: yeah sounds good, lets do that for now 👍

might be nice to pop a comment at the top of the utils pyproject.toml saying "this file was generated based on notification_utils" or similar

quis: My script does that automatically 😄 :

image

***

I’ve kept `pyproject.toml` in source control now and squashed that change into 72de27e

leohemsted: 🧙 🪄

leohemsted (approved):

PR #243, opened 2024-07-12
Drop Python 3.7 support

quis:
Python 3.7 is now end-of-life

The last security-only release of Python 3.7 was on 6th June 2023: https://peps.python.org/pep-0537/#schedule-last-security-only-release

> The code base for 3.7 is now frozen and no further updates will be provided nor issues of any kind will be accepted on the bug tracker.

I’ll release a new version of the client once this and https://github.com/alphagov/notifications-python-client/pull/242 are merged

***

Doing this now so that the tests can move to a newer version of `requests` which doesn’t suport Python 3.7 any more.


klssmith:
There are also mentions of Python 3.7 in `setup.py` which could be removed

quis:
@klssmith done and squashed

PR #244, opened 2024-07-31
Add Makefile command to bump utils version

quis:
We have this `make` command in most of our other repos, for example the admin app: https://github.com/alphagov/notifications-admin/blob/main/Makefile#L85-L87

***

Depends on:
- [x] https://github.com/alphagov/notifications-utils/pull/1137


leohemsted (approved):

tombye (approved):
Looks good to me 👍🏻

PR #284, opened 2026-09-21
Add type checking with mypy

quis:
Add the dependencies and Make commands per the our other repos.

Config needs to be in `mypy.ini` because we haven’t moved this repo to use `pyproject.toml` yet.

Running `mypy` gives:
> Found 6 errors in 2 files (checked 6 source files)

***

This PR fixes those errors, then turns on type checking in CI so more can’t get introduced.


joybytes (approved):

risicle (approved):

risicle:
That iterator method is wild - just looks for anything that looks like a uuid.

risicle:
This could probably do with a point release.

